PR #531, opened 2022-10-28
feat(tokens): add global z-index and duration tokens

Niznikr:
## Summary

- Add global z-index and duration tokens
- z-index values go from 100 - 900 (modal: 500, popover/dropdown: 700, tooltip: 800, toast/snackbar: 900)
- Add story to illustrate z-index hierarchy


NazimHAli:
Looking awesome 🎉

PR #583, opened 2022-11-22
feat: add alias token story

Niznikr:
## Summary

Add story for alias tokens and update global tokens story to match. Token names can be copied for easy access.

Screen Shot 2022-11-22 at 10 16 05 AM


chasedarkly (approved):
Love it, looks awesome

PR #603, opened 2022-12-06
feat: auto-collapse Navigation to prevent scroll

kwatkins-ld:
## Summary

Updating `Navigation` to auto-collapse if the rendered `nav` element begins to scroll horizontally. This replaces the old behavior of collapsing when the viewport was less than 740px.

### DIFF
--- a/packages/navigation/src/NavigationMenuDropdown.tsx
+++ b/packages/navigation/src/NavigationMenuDropdown.tsx
@@ -1,66 +1,43 @@
 import type { CollectionBase } from '@react-types/shared';
+import type { ChangeEvent } from 'react';
 
-import { Chip } from '@launchpad-ui/chip';
-import { Dropdown, DropdownButton } from '@launchpad-ui/dropdown';
-import { Menu, MenuItem } from '@launchpad-ui/menu';
+import { Select } from '@launchpad-ui/form';
 import { useListState } from '@react-stately/list';
-import { useEffect, useRef, useState } from 'react';
-import { NavLink } from 'react-router-dom';
+import { useLocation, useNavigate } from 'react-router-dom';
 
 import { titlecase } from './utils';
 
-type NavigationMenuDropdownProps<T extends object> = CollectionBase<T> & {
-  title: string;
-};
+type NavigationMenuDropdownProps<T extends object> = CollectionBase<T>;
 
-const NavigationMenuDropdown = <T extends object>(props: NavigationMenuDropdownProps<T>) => {
+const NavigationMenuDropdown = <T extends object>(props: CollectionBase<T>) => {
   const state = useListState(props);
-  const ref = useRef<HTMLDivElement>(null);
-  // Track the width of the dropdown button so that we can set the menu content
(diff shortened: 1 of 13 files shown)


packages/navigation/src/utils.ts
```diff
@@ -1,28 +1,8 @@
-import { useEffect, useState } from 'react';
-
 const titlecase = (str: string) => {
   return str
     .toString()
     .toLowerCase()
     .replace(/\b([a-z])/g, (ch) => ch.toUpperCase());
 };
 
-const useMediaQuery = (query: string) => {
```
kwatkins-ld, line 10: This became dead code once we removed the `740px` check.

packages/navigation/src/NavigationMenuDropdown.tsx
```diff
@@ -14,31 +15,52 @@ type NavigationMenuDropdownProps<T extends object> = CollectionBase<T> & {
 
 const NavigationMenuDropdown = <T extends object>(props: NavigationMenuDropdownProps<T>) => {
   const state = useListState(props);
```
kwatkins-ld, line 17: This is approach seems necessary because:

* There doesn't seem to be a way to configure a dropdown to set each menu item to the dropdown button's width
* The dropdown button gets cloned by dropdown before rendering as a popover, so we can't set a ref on `DropdownButton` from the outside

apucacao: Is this related to the collapse mechanism?

apucacao: Ignore me. We handled it in chat.

packages/navigation/src/NavigationMenuDropdown.tsx
```diff
@@ -14,31 +16,56 @@ type NavigationMenuDropdownProps<T extends object> = CollectionBase<T> & {
...
+    const button = ref.current?.querySelector('button');
+
+    if (!button) {
+      return;
+    }
+
+    setDropdownButtonWidth(button.getBoundingClientRect().width);
+  }, [setDropdownButtonWidth]);
 
   return (
```
kwatkins-ld, line 36: "Hide Whitespace" helps a bit here:

* Wrap the rendered `Dropdown` in a new `div` with a ref, and add a `placement` prop
* Wrap the `onClick` prop, so that in addition to the event param we can add an additional `{ collapsed: true }` state param

kwatkins-ld:
@chasedarkly I think this is now ready for review. I wasn't sure if removing the `on-hold` label would cause confusion so I opted to leave it 😅

packages/navigation/src/styles/Navigation.module.css
```diff
@@ -12,15 +12,20 @@
 
 .Navigation .Nav--primary {
   padding: 0 2.5rem;
```
kwatkins-ld, line 14: The background-color and box-shadow are being removed, with the expectation that consumers manage this going forward.

chasedarkly (approved):

PR #615, opened 2022-12-13
feat(form,tokens): improve form theming

chasedarkly:
## Summary
- Added `*-field-error` tokens to apply theming consistently across fields in an error state

### DIFF
--- a/packages/form/src/styles/Form.module.css
+++ b/packages/form/src/styles/Form.module.css
@@ -70,6 +70,13 @@ select.formInput {
   border-radius: var(--lp-border-radius-regular) 0 0 var(--lp-border-radius-regular);
 }
 
+.isInvalid .formInput,
+.form .isInvalid :global(.Select-control),
+.form .isInvalid :global(.CustomSelect) > div,
+.form .isInvalid .formInput {
+  border-color: var(--lp-color-border-field-error);
+}
+
 .suffixContainer .formInput:focus {
   box-shadow: none;
 }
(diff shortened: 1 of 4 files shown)


packages/form/src/styles/Form.module.css
```diff
@@ -70,6 +70,13 @@ select.formInput {
   border-radius: var(--lp-border-radius-regular) 0 0 var(--lp-border-radius-regular);
 }
 
+.isInvalid .formInput,
+.form .isInvalid :global(.Select-control),
+.form .isInvalid :global(.CustomSelect) > div,
+.form .isInvalid .formInput {
```
kwatkins-ld, line 76: I'm guessing there is a good reason, but at a glance it seems like either the first selector or this last one could be omitted? Is the first selector being overridden due to specificity in some cases?

chasedarkly: Although the diff looks like I added all of these - I only added `.isInvalid .formInput`, but when I added it, I had to move it higher in the file to avoid a specificity cascade lint error. The reason I added `.isInvalid .formInput` is because I wanted to fix a storybook story that correctly shows a field error but wasn't applying an invalid state border color to the example. It's possible we could now omit `.form .isInvalid .formInput`, but I don't want to break anything upstream that was relying on this higher specificity.

kwatkins-ld: Ah, that makes sense. Thanks!

kwatkins-ld (approved):

PR #621, opened 2022-12-15
chore: add build:watch tooling to enable local linking

kwatkins-ld:
## Summary

Adding support for linking LaunchPad locally.

### DIFF
--- a/.plop/templates/component/package.json.hbs
+++ b/.plop/templates/component/package.json.hbs
@@ -27,6 +27,7 @@
   "source": "src/index.ts",
   "scripts": {
     "build": "vite build -c ../../vite.config.ts && tsc --project tsconfig.build.json",
+    "build:watch": "concurrently -n vite,tsc -c blue,green 'vite build --watch -c ../../vite.config.ts' 'tsc --preserveWatchOutput --project tsconfig.build.json --watch'",
     "clean": "rm -rf dist",
     "lint": "eslint '**/*.{ts,tsx,js}' && stylelint '**/*.css' --ignore-path ../../.stylelintignore",
     "test": "vitest run --coverage"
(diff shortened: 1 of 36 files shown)


.changeset/big-fishes-rest.md
```diff
@@ -0,0 +1,5 @@
+---
+'@launchpad-ui/core': patch
+---
+
+Adding build:watch tooling to enable local linking
```
chasedarkly, line 5: You can delete this changeset file - since nothing new is being shipped to consumers we don't need to bump versions

kwatkins-ld: Ah, thanks! I thought it was required but that makes sense.

.plop/templates/component/package.json.hbs
```diff
@@ -27,6 +27,7 @@
   "source": "src/index.ts",
   "scripts": {
     "build": "vite build -c ../../vite.config.ts && tsc --project tsconfig.build.json",
```
kwatkins-ld, line 29: @chasedarkly I almost added `concurrently` to `devDependencies` in all the package-level `package.json` files (because it's required for the new `build:watch` commands in each package), but based on how we treat `eslint`, `vite`, `tsc`, etc. it seemed that the convention is to let those tools be managed at the top-level.

Does that sound right to you?

chasedarkly: Sounds right to me 👍

kwatkins-ld:
Closing this for now, there are some subtle problems we'll need to solve for this to be viable:

* There seem to be race conditions between `vite` and `tsc` running concurrently: in some cases the build output would not include `require` statements to stylesheets imported by components
* The top-level package script `build:watch` that runs the same package script in each package doesn't appear to work in all cases: the `@launchpad-ui/button` package for example never emits new build output

PR #648, opened 2023-01-12
refactor(banner): remove dismissible prop in favor of onDismiss prop

traciklopez:
## Summary

Fix bugs where `dismissible` props were passed to the banner component without an `onDismiss` function. This results in the banner _not_ being dismissed when clicked.


chasedarkly (approved):
perfect - nice work! thanks

PR #661, opened 2023-01-20
chore: improve Navigation collapse behavior

kwatkins-ld:
## Summary

* Use `useLayoutEffect` when choosing to collapse the `Navigation`, to prevent a visible render which causes a flicker as the tabs get replaced with a `Select`
* Refactor the collapsed implementation to render a `Dropdown` instead of a `Select` (better a11y posture)
* Wraps the collapsed navigation content in a `nav` element with an `aria-label` attribute, to match the non-collapsed implementation
* Update the styling of the tab labels based on design feedback


packages/navigation/src/Navigation.tsx
```diff
@@ -47,7 +47,7 @@ const Navigation = <T extends object>(props: NavigationProps<T>) => {
     });
   }, [wrapperRef, itemListRef, setCollapse]);
```
kwatkins-ld, line 49: This prevents the flicker that would otherwise occur when the `Navigation` renders as tabs before collapsing down to the `Select`.

packages/navigation/src/NavigationMenuSelect.tsx
```diff
@@ -28,24 +49,16 @@ const NavigationMenuSelect = <T extends object>(props: CollectionBase<T>) => {
   );
 
   return (
```
kwatkins-ld, line 51: This diff is a bit complicated:

* Add `nav` with `aria-label` around the `Select`
* Remove the `selected` prop and associated logic (lifted up to the new `useState` above, and set as `value` on the `Select` instead)

packages/navigation/src/NavigationMenuDropdown.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+type NavigationMenuDropdownProps<T extends object> = CollectionBase<T> & {
+  kind: NavProps['kind'];
+};
+
+const NavigationMenuDropdown = <T extends object>(props: NavigationMenuDropdownProps<T>) => {
+  const state = useListState(props);
+  const { pathname } = useLocation();
+  const buttonRef = useRef<HTMLButtonElement>(null);
+  const [buttonWidth, setButtonWidth] = useState<string | undefined>(undefined);
```
kwatkins-ld, line 23: As the button text changes, the width of the button also changes... we're tracking the state to use it as a `minWidth` on each menu item so that the menu takes is at least as wide as the button (but will grow wider if necessary)

packages/navigation/src/styles/Navigation.module.css
```diff
@@ -57,8 +57,7 @@
   align-items: center;
```
kwatkins-ld, line 57: Lots of changes in this file, the gist is:

* The tabs navigation are going from 24px left/right padding down to `15`.
* The bottom border is moving from the `a` to the `span` that has the navigation item text, so that it more closely hugs the text width without reducing click target area

chasedarkly: Why 15 instead of the 8-based 16?

packages/navigation/src/styles/Navigation.module.css
```diff
@@ -176,3 +189,7 @@
 .NavItem-tooltip {
   text-align: center;
 }
+
```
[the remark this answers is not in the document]
chasedarkly: You could try targeting `.NavigationMenuItem svg` and see if that works, otherwise this seems reasonable. We could also modify the API so that `icon` takes a ReactNode, which you can pass a class to. Then, internally in MenuItem we'd clone the element and add the `Menu-item-icon` class as well.

chasedarkly (approved):

packages/navigation/src/styles/Navigation.module.css
kwatkins-ld: We talked through this, landed on `12px` for left/right padding in the `a` element, `4` for left/right padding in the `span` for the tab label (`16px` in total on either side)

PR #671, opened 2023-01-30
feat(select): add Select package

chasedarkly:
## Summary
Look at the Storybook examples and rendered output to see how this works.


packages/select/src/__tests__/Select.cy.tsx
```diff
@@ -0,0 +1,25 @@
...
+        <Select label="Fruit" selectionMode="single" items={FRUIT}>
+          {(item) => <Item>{item.name}</Item>}
+        </Select>
+      );
+    });
+
+    it('renders', () => {
+      cy.getByTestId('select').should('be.visible');
+    });
+
```
kwatkins-ld, line 20: I think it'd be great to have some additional coverage here for things like section groups, custom triggers, custom item rendering, etc.

We don't need to address that in this PR, and it might also be a good idea for someone else (I'll volunteer, but it doesn't have to be me) to write the additional tests to cross-pollinate knowledge of this work. What do you think?

chasedarkly: Check out `Select.spec.tsx` -- I initially had tests in the cy file, but the spec file is the one that has coverage thresholds

chasedarkly: I didn't test sections explicitly, though. I definitely can, but since the rendered difference with sectioned items vs items is basically just the addition of a nested ul/li structure, I prioritized other tests

kwatkins-ld: Ah I see, hadn't made my way to the unit tests yet... I think this level of coverage is good, and agree that there are probably higher priorities.

kwatkins-ld (approved):

PR #678, opened 2023-02-03
feat(markdown): add optional "children" prop to permit custom rendering logic in component consumer

ldhenry:
## Summary

This PR adds a new a new `childern` property that will give component consumers the ability to add custom rendering if needed.

Here is an example of how the component can be used:

```tsx
<Markdown
        source={SAMPLE_MARKDOWN}
        allowedTags={['a', '#text', 'p', 'li', 'ol', 'ul', 'b', 'strong', 'i', 'em', 'br', 'code']}
      >
        {parsedMarkdown => parse(parsedMarkdown).replace(customReplaceFn)}
</Markdown>
```

I confirmed that including `html-react-parser` as a dev dependency does not increase the bundle size but still permits a successful storybook build.

More details on how we landed on this implementation can be found [here](https://launchdarkly.slack.com/archives/C01TNPM36U8/p1675459059487959)

## Testing approaches

- New story
- New test
- Existing tests


packages/markdown/package.json
```diff
@@ -34,6 +34,7 @@
   "dependencies": {
     "@launchpad-ui/tokens": "workspace:~",
     "classix": "2.1.17",
```
kwatkins-ld, line 36: This will add almost [10k](https://bundlephobia.com/package/html-react-parser@3.0.8) to the bundle-size for this package, and may end up causing confusion given that we're also using `marked` ... I don't have much context for this problem, but I'm curious if we tried using the [`renderer` option from `marked`](https://marked.js.org/using_pro#renderer) as a solution?

I imagine that we'd need a way to thread that in from consumers, e.g. changes to the props surface area for the `Markdown` component, but that might allow us to avoid introducing a new parsing library and instead of a new `smartLinks` prop we'd end up with a way to more generically customize rendering behavior.

ldhenry: Moved discussion to [slack](https://launchdarkly.slack.com/archives/C01TNPM36U8/p1675459228682109?thread_ts=1675459059.487959&cid=C01TNPM36U8). I'll update this thread when we come to a conclusion.

chasedarkly (approved):
Looks good, thanks!

PR #683, opened 2023-02-07
feat(drawer): add cancel icon toggle to Drawer

tvarney13:
## Summary

In the New Flag Creation Experience I had to add this CSS hack to make the cancel icon go away in Drawer:

```
.entity-creation__drawer > div > div > .IconButton {
  display: none;
}
```

We did this because this Drawer had several bits of custom UI for breaking out of it, and the icon was occupying space in the header of the drawer.

image

To avoid doing such hacks in the future, Squonk wanted to add a toggle to the `Drawer`. It's a boolean flag that controls whether the `IconButton` appears in the DOM or not.

## Testing approaches

Added unit test coverage.

### DIFF
--- a/packages/drawer/__tests__/Drawer.spec.tsx
+++ b/packages/drawer/__tests__/Drawer.spec.tsx
@@ -13,6 +13,7 @@ describe('Drawer', () => {
       </Drawer>
     );
     expect(await screen.findByRole('dialog')).toBeInTheDocument();
+    expect(await screen.findByTestId('drawer-close-button')).toBeInTheDocument();
   });
 
   it('calls onCancel when escape key is pressed', async () => {
--- a/packages/drawer/__tests__/Drawer.spec.tsx
+++ b/packages/drawer/__tests__/Drawer.spec.tsx
@@ -28,4 +28,14 @@ describe('Drawer', () => {
 
     expect(spy).toHaveBeenCalledTimes(1);
   });
+
+  it('hides cancel button when hideCancel is passed in', async () => {
+    render(
+      <Drawer hideCancel>
+        <DrawerHeader>Drawer</DrawerHeader>
+      </Drawer>
+    );
+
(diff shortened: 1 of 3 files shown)


packages/drawer/__tests__/Drawer.spec.tsx
```diff
@@ -13,6 +13,7 @@ describe('Drawer', () => {
       </Drawer>
     );
     expect(await screen.findByRole('dialog')).toBeInTheDocument();
+    expect(await screen.findByTestId('drawer-close-button')).toBeInTheDocument();
```
tvarney13, line 16: Wanted to test the case where `hideCancel` isn't explicitly passed in as well.

packages/drawer/src/Drawer.tsx
```diff
@@ -49,6 +50,7 @@ const Drawer = ({
   onCancel,
   size = 'small',
   'data-test-id': testId = 'drawer',
```
kwatkins-ld, line 52: I wonder if this might be easier to reason about if we inverted the condition, e.g. a `showCancelButton` prop with a default of `true`?

I understand framing it as a deviation from the default behavior, but in general I think negative conditions are harder to parse (that's personal/subjective though).

tvarney13: That's actually what I started with! However the ergonomics of `showCancel={false}` in the tags felt off to me vs a more declarative `hideCancel` - like you're conveying to the code reader what the drawer is doing.

tvarney13: I admit it's weird to reason about in the code but I think it makes enough of a positive difference to the interface to offset that confusion.

chasedarkly: Looks like there is precedence for `hide*` style props in LaunchPad, too

chasedarkly (approved):

PR #686, opened 2023-02-09
fix(modal,drawer): update overlay bg colors

chasedarkly:
## Summary
Design has asked the modal and drawer overlay backgrounds be updated.

### DIFF
--- a/packages/modal/src/styles/Modal.module.css
+++ b/packages/modal/src/styles/Modal.module.css
@@ -14,7 +14,7 @@
 .overlay {
   position: fixed;
   inset: 0;
-  background-color: hsl(0 0% 20% / 0.5);
+  background-color: rgb(40 40 40 / 0.5);
   display: flex;
   align-items: flex-end;
   justify-content: center;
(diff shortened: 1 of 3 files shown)


packages/modal/src/styles/Modal.module.css
```diff
@@ -14,13 +14,13 @@
...
   inset: 0;
-  background-color: hsl(0 0% 20% / 0.5);
+  background-color: rgb(40 40 40 / 0.5);
   display: flex;
   align-items: flex-end;
   justify-content: center;
 
   :root[data-theme='dark'] & {
-    background-color: rgb(18 18 18 / 0.75);
+    background-color: rgb(25 25 25 / 0.75);
```
kwatkins-ld, line 23: Just to confirm: the default theme backgrounds are the same for drawer/modal, but the dark them has a slightly different alpha value (`0.75` here in the modal, vs `0.7` in the drawer)?

chasedarkly: Good catch! Fixing that

chasedarkly: Fixed.

kwatkins-ld (approved):

PR #692, opened 2023-02-15
chore(deps): update framer motion from 7.6.1 to 9.0.3

chasedarkly:
## Summary
Followed along with the [upgrade guide for framer-motion](https://www.framer.com/motion/guide-upgrade/) to determine that we can safely update framer-motion to 9.0.3 from 7.6.1. Let's release this as a standalone version upgrade to LaunchPad so if something does pop up in Gonfalon, we can revert this in isolation. Gonfalon _also_ imports `framer-motion`, and I tested there that upgrading to 9.0.3 does not result in any test failures.

### DIFF
--- a/.changeset/chilly-rice-doubt.md
+++ b/.changeset/chilly-rice-doubt.md
@@ -1,0 +1,18 @@
+---
+'@launchpad-ui/clipboard': patch
+'@launchpad-ui/markdown': patch
+'@launchpad-ui/snackbar': patch
+'@launchpad-ui/popover': patch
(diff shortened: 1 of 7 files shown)


apucacao (approved):
Do we have a sense of bundle size changes?

chasedarkly:
@apucacao According to bundlephobia, [9.0.3](https://bundlephobia.com/package/framer-motion@9.0.3) is slightly smaller than [7.6.1](https://bundlephobia.com/package/framer-motion@7.6.1)

PR #694, opened 2023-02-16
fix: update button styles to match Figma

jlzych:
## Summary

I updated the button styles to match Figma.

Screenshot 2023-02-15 at 4 15 19 PM

## Testing approaches

I just tested visually.

One thing to note is that I _think_ Launchpad isn't loading Inter, which makes developing in Launchpad tough since the font will always be different. I installed Inter locally while making these changes which I think fixed it, but it's hard for me to know for sure since Inter and the fallback font (apple system font? SF Pro?) are pretty similar. So if we could Inter in Storybook while developing it would help us stay closer to how everything is styled in gonfalon.

### DIFF
--- a/.changeset/rich-ways-remember.md
+++ b/.changeset/rich-ways-remember.md
@@ -1,0 +1,7 @@
+---
+'@launchpad-ui/split-button': patch
+'@launchpad-ui/button': patch
+'@launchpad-ui/core': patch
+---
+
+[Button] Adjust sizing to match design system guidelines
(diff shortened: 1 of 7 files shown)


chasedarkly:
Please follow the contribution docs to generate a changelog file: https://github.com/launchdarkly/launchpad-ui/blob/main/CONTRIBUTING.md

chasedarkly:
Confirmed with @jlzych that these changes affect the size (height/width) of buttons, and I brought up my concern that much of our UI in Gonfalon containing buttons was built (in many cases) around the current sizing, and changing it will likely cause some visual regressions we'd need to address in Gonfalon. We are going to look into this tomorrow to see if we can de-risk or come up with a good way to test this.

chasedarkly (approved):

PR #707, opened 2023-02-28
feat(select): add better select contract with SingleSelect and MultiSelect

chasedarkly:
## Summary
This looks big, but it basically just splits out what Select was doing into two components: `MultiSelect` and `SingleSelect`. I thought we could get away with sharing one component, but the downstream implications on type safety and sharing contracts proved too burdensome when trying to expose methods that better fit the "single selected value" experience vs the "multiple selected value" experience. I was aware of this already, and when @vroske-ld pointed out that she didn't love  that when building a single select component we have to access the selected key like `selectedKeys[0]`.

By making these into two separate components, we had to duplicate _some_ code, so the package size is slightly larger - but we were also able to remove some code that existed purely to fork between single/multiple behavior.


packages/select/src/__tests__/MultiSelect.cy.tsx
```diff
@@ -1,16 +1,16 @@
...
+import { MultiSelect } from '../';
 
 import { FRUIT } from './constants';
 
 describe('Select', () => {
   describe('with default single select trigger', () => {
     beforeEach(() => {
       cy.mount(
-        <Select label="Fruit" selectionMode="single" items={FRUIT}>
+        <MultiSelect label="Fruit" items={FRUIT}>
```
vroske-ld, line 11: this seems like it should be `SingleSelect`?

packages/select/stories/SingleSelect.stories.tsx
```diff
@@ -0,0 +1,192 @@
...
+
+type Story = StoryObj<typeof SingleSelect>;
+
+export const Basic: Story = {
+  render: () => {
+    const Component = () => {
+      return (
+        <SingleSelect
+          label="Fruit"
+          defaultItems={FRUIT}
```
vroske-ld, line 32: let's think more about the `defaultItems` vs `items` thing. personally i'd have a hard time keeping this straight

packages/select/src/__tests__/examples.tsx
```diff
@@ -1,10 +1,11 @@
 import type { DummyItem } from './constants';
```
vroske-ld, line 1: this is minor, but i usually see stuff like this being referred to as 'fixtures', since it seems this is used in a test. As opposed to like an example test, or a story for Storybook.

packages/select/src/__tests__/SingleSelect.spec.tsx
```diff
@@ -0,0 +1,98 @@
...
+          )}
+          items={FRUIT}
+        >
+          {(item) => <Item>{item.name}</Item>}
+        </SingleSelect>
+      );
+
+      expect(screen.getByTestId('select')).toBeVisible();
+
+      await user.click(screen.getByTestId('select-trigger'));
```
vroske-ld, line 67: we probably want to consider duplicating these same test scenarios, but firing with keyboard events instead

packages/select/src/SingleSelect/SingleSelectTrigger.tsx
```diff
@@ -0,0 +1,58 @@
...
+  children?: (state: SingleSelectTriggerChildrenState<T>) => ReactNode;
+};
+
+const SingleSelectTrigger = <T extends object>(props: SingleSelectTriggerProps<T>) => {
+  const { state, triggerProps, valueProps, triggerRef, children } = props;
+
+  const getRenderedSelected = () => {
+    const item = state.selectedItem;
+
+    if (!item) return 'Select option';
```
vroske-ld, line 26: i wonder if this placeholder text should be expose as a prop?

packages/select/src/SingleSelect/useSingleSelectState.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  return {
+    ...listState,
+    ...triggerState,
+    close() {
+      triggerState.close();
+    },
+    open() {
+      // Don't open if the collection is empty.
+      if (listState.collection.size !== 0) {
```
vroske-ld, line 51: i see a lot of checks and conditionals like this, we should try to cover these scenarios with tests

vroske-ld (commented):
these components are so complex i fully support `SingleSelect` and `MultiSelect` as distinct entities. 😅

vroske-ld (approved):

packages/select/src/SingleSelect/types.ts
```diff
@@ -0,0 +1,12 @@
+import type { Key } from 'react';
+
+type SingleSelection = {
+  /** The currently selected keys in the collection (controlled). */
+  selectedKey?: Key;
+  /** The initial selected keys in the collection (uncontrolled). */
+  defaultSelectedKey?: Key;
+  /** Handler that is called when the selection changes. */
+  onSelectionChange?: (key: Key) => any;
+};
```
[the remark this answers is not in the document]
chasedarkly: in this case, the `useSingleSelectListState` component internally converts the `react-stately` collections API interface  that only has methods for multi-select into single-select, by essentially wrapping the value in an array.

packages/select/src/styles/Select.module.css
```diff
@@ -167,3 +168,25 @@
...
+
+  input {
+    width: 100%;
+    border: none;
+    font-size: 1.6rem;
+    padding: 1.6rem 1.6rem 1.6rem 4rem;
+
+    &:focus-visible {
+      outline: none;
+    }
```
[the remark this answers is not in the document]
chasedarkly: Currently, when the popover is open, the input field is always focused, and the list has virtual focus. We don't have a strong focus indicator currently, it just shows the search field as active

packages/select/src/SingleSelect/SingleSelectTrigger.tsx
chasedarkly: Oh yes, definitely should be.

PR #708, opened 2023-02-28
feat(snackbar): add optional footer to Snackbar

jagarnica:
## Summary

This adds an optional footer in snackbars controlled via the prop: "footer"

Example:
```js
dispatchSnackbar({
 kind: 'error',
 description: 'There was an issue processing your request',
 footer: <div>View More</div>,
});
```
Screenshot 2023-02-28 at 4 07 35 PM


chasedarkly:
Shoot - we decided against this pattern, we talked about that last week, sorry if there was confusion. I confirmed with Linda yesterday as well to make sure we still don't want this. Unless there is some info i'm missing, we can close this I think.

jagarnica:
> Shoot - we decided against this pattern, we talked about that last week, sorry if there was confusion. I confirmed with Linda yesterday as well to make sure we still don't want this. Unless there is some info i'm missing, we can close this I think.

All good! I'll close the one.

PR #710, opened 2023-03-01
feat(modal): Add support for larger modals

ldhenry:
## Summary

This PR adds support for a "large" modal size. While it may make more since to call it "wide" since it is simply wider, I decided to go with "large" to match the "small" size that already exists.

It looks like we may have supported the `auto` size in the past. If you prefer, we could revert back to that approach to give component consumers the ability to set their own modal widths with CSS.

CleanShot 2023-03-01 at 1 47 55

## Testing approaches

- manual
- [added new storybook story](https://626696a2018c1f004a1cde86-aollkbidvd.chromatic.com/?path=/story/components-modal--large)


.changeset/kind-colts-pay.md
```diff
@@ -0,0 +1,5 @@
+---
+'@launchpad-ui/modal': minor
```
chasedarkly, line 2:
```suggestion
'@launchpad-ui/modal': minor
'@launchpad-ui/core': minor
```

.changeset/kind-colts-pay.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+Add support for large size.
```
chasedarkly, line 5:
```suggestion
[Modal] Add `large` variant
```

packages/modal/stories/Modal.stories.tsx
```diff
@@ -147,6 +147,30 @@ export const Small: Story = {
...
+    return (
+      <Modal size="large">
+        <ModalHeader title="Heading" />
+        <ModalBody>
+          <p>
+            The <code>large</code> modal is only appropriate when the contents of the modal
+            necessitate the extra width.
+          </p>
+          <p>
+            <strong>This size should be used sparingly.</strong>
```
chasedarkly, line 161: Love the use of the example to add usage reccs!

chasedarkly (approved):

PR #721, opened 2023-03-07
refactor(select): expose SelectItem and SelectSection

chasedarkly:
## Summary
Based on some conversation with the team, we decided not to require consumers to use `react-stately`s `Item` and `Section` components directly, and instead expose named components through LaunchPad. These components in this case are `SelectItem` and `SelectSection`. 

I also added the `as` prop to `SelectItem`, which allows us to render selected items as links if needed, like in the case of the refactored Navigation component in this PR: https://github.com/launchdarkly/launchpad-ui/pull/696/files#diff-622d77675b453717536c6747bff490028f4b3ab94fd52c74f38dec365d91c1ffR23

### DIFF
--- a/packages/select/src/MultiSelect/MultiSelect.tsx
+++ b/packages/select/src/MultiSelect/MultiSelect.tsx
@@ -40,7 +40,7 @@ const MultiSelect = <T extends object>(props: MultiSelectProps<T>) => {
   const filterInputRef = useRef<HTMLInputElement>(null);
   const triggerRef = useRef<HTMLButtonElement>(null);
   const popoverRef = useRef<HTMLDivElement>(null);
-  const listBoxRef = useRef<HTMLUListElement>(null);
+  const listBoxRef = useRef<HTMLDivElement>(null);
 
   const state = useMultiSelectState(props);
 
--- a/packages/select/src/SelectItem.tsx
+++ b/packages/select/src/SelectItem.tsx
@@ -8,7 +8,7 @@ type SelectItemProps<T extends object, P extends ElementType> = ItemProps<T> & {
   [key: string]: any;
 };
 
-const SelectItem = <T extends object, P extends ElementType = 'li'>(
+const SelectItem = <T extends object, P extends ElementType = 'div'>(
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   _props: SelectItemProps<T, P> & Omit<ComponentPropsWithoutRef<P>, keyof SelectItemProps<T, P>>
 ) => {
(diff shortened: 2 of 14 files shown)


packages/select/src/MultiSelect/MultiSelect.tsx
```diff
@@ -40,7 +40,7 @@ const MultiSelect = <T extends object>(props: MultiSelectProps<T>) => {
   const filterInputRef = useRef<HTMLInputElement>(null);
   const triggerRef = useRef<HTMLButtonElement>(null);
   const popoverRef = useRef<HTMLDivElement>(null);
-  const listBoxRef = useRef<HTMLUListElement>(null);
+  const listBoxRef = useRef<HTMLDivElement>(null);
```
chasedarkly, line 43: I checked with a screenreader to confirm that changing the list from `ul/li` to `div`s doesn't affect accessibility / voiceover output, since we have the correct roles on the list and list items.

packages/select/src/SelectItem.tsx
```diff
@@ -0,0 +1,44 @@
+import type { SharedSelectState } from './types';
+import type { ItemProps } from '@react-types/shared';
+import type { ComponentPropsWithoutRef, ElementType } from 'react';
+
```
chasedarkly, line 4: For anyone thoroughly confused by this -- `react-stately`'s collections use React components in a _slightly_ unique way. This is effectively a pass-through component that exposes `getCollectionNode` to be used internally by `react-stately` and `react-aria`.

packages/select/src/MultiSelect/MultiSelect.tsx
apucacao: Bonus points: record a little loom and share it in #ask-frontend.

packages/select/src/SelectItem.tsx
```diff
@@ -0,0 +1,44 @@
...
+type SelectItemProps<T extends object, P extends ElementType> = ItemProps<T> & {
+  as?: P;
+  onClick?: (_e: MouseEvent, state: SharedSelectState) => void;
+  [key: string]: any;
+};
+
+const SelectItem = <T extends object, P extends ElementType = 'li'>(
```
apucacao, line 11: style thing: why did you go with an assignment expression instead of a function declaration?

apucacao (approved):

packages/select/src/SelectItem.tsx
chasedarkly: I'm not super picky on one vs the other but mainly because I like consistency where possible -- e.g. everything can be an assignment expression, but not everything can be a function declaration. Which do you prefer?

chasedarkly: PS you made me notice a bug by commenting on this line 😂 `ElementType = 'li'` should be `ElementType = 'div'`

apucacao: :success:

I tend to prefer `function` when defining something a function because it maps nicely to a human description of the code, but that's not always "convenient" so I'm also not super picky. In this case, I was curious because the type parameter TypeScript syntax for arrow functions always takes me a second to understand. 😅

PR #754, opened 2023-03-10
fix: add release-skip workflows

chasedarkly:


.github/workflows/chromatic.yml
```diff
@@ -1,9 +1,6 @@
 name: Chromatic
 
-on:
-  push:
-    branches-ignore:
-      - 'changeset-release/**'
+on: push
```
chasedarkly, line 3: I don't think there is actually any reason to skip this - we can and probably should just run chromatic on changeset-release branches

.github/workflows/package-size.yml
```diff
@@ -1,9 +1,6 @@
 name: Compressed Size
 
-on:
-  pull_request:
-    branches-ignore:
-      - 'changeset-release/**'
+on: pull_request
```
chasedarkly, line 3: Same here - I'm not sure why we should even prevent this from running.

.github/workflows/verify.yml
```diff
@@ -4,10 +4,6 @@ on:
   pull_request:
     branches-ignore:
       - 'changeset-release/**'
-  push:
-    branches:
-      - main
-      - 'release/**'
```
chasedarkly, line 10: This was actually the culprit for why status checks weren't running. There is really no reason to filter on this.

kparkinson-ld (approved):

PR #768, opened 2023-03-13
chore: update renovate to only update range deps

chasedarkly:
We have several packages set to pinned dependency versions, and it makes sense to update these manually where possible instead of having renovate bot suggest we update them. These versions are also matched to ones in the downstream consumer app to minimize bundle size.


stinachen (approved):
⭐️

PR #782, opened 2023-03-16
feat: new `action` prop for tags, improve `tiny` button styles, and rename `default` button kind to `basic`

chasedarkly:
## Summary
[Button]:
- Rename `default` kind to `basic`
- rename size variants for `ButtonGroup`
- Change height of tiny button to 20px to match tiny tags

[Tag]:
- Remove `onAction` and `actionLabel` props in favor of a new `action` prop which accepts a render element
- Improve styles and add stories

### DIFF
--- a/packages/button/src/IconButton.tsx
+++ b/packages/button/src/IconButton.tsx
@@ -14,7 +14,7 @@ import { isValidElement, cloneElement, forwardRef, memo } from 'react';
 import './styles/Button.css';
 
 type IconButtonProps = ButtonHTMLAttributes<HTMLButtonElement> & {
-  kind?: 'default' | 'primary' | 'destructive' | 'minimal' | 'close';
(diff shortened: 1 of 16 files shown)


packages/button/src/IconButton.tsx
```diff
@@ -14,7 +14,7 @@ import { isValidElement, cloneElement, forwardRef, memo } from 'react';
 import './styles/Button.css';
 
 type IconButtonProps = ButtonHTMLAttributes<HTMLButtonElement> & {
-  kind?: 'default' | 'primary' | 'destructive' | 'minimal' | 'close';
+  kind?: 'basic' | 'primary' | 'destructive' | 'minimal' | 'close';
```
chasedarkly, line 17: The "default" kind was only the actual default for Buttons, so `basic` is a better name.

apucacao (approved):

PR #789, opened 2023-03-20
fix(filter): improve clear icon sizing

chasedarkly:
## Summary

BEFORE:
Screenshot 2023-03-20 at 12 07 31 PM

AFTER:
Screenshot 2023-03-20 at 12 07 15 PM


packages/select/src/__tests__/examples.tsx
```diff
@@ -2,40 +2,33 @@ import type { DummyItem } from './constants';
...
 
-import { Tooltip } from '@launchpad-ui/tooltip';
+import { IconButton } from '@launchpad-ui/button';
+import { Edit } from '@launchpad-ui/icons';
+import { TagGroup, TagItem } from '@launchpad-ui/tag';
 
 const CustomMultiSelectTrigger = (props: MultiSelectTriggerProps<DummyItem>) => {
   const { state, triggerProps, triggerRef } = props;
 
+  console.log(state.selectedItems);
```
kwatkins-ld, line 12: Guessing we wanna remove this.

chasedarkly: These files crept in, shouldn't be in the PR. Thanks for flagging!

PR #798, opened 2023-03-28
fix(toggle): rework toggle to use label around component

chasedarkly:
## Summary
- Rework toggle to rely more on the `react-aria`/`react-stately` hooks under the hood, similar to React Spectrum's implementation.
- Support controlled _or_ uncontrolled behavior, and show this in stories
- Remove label text props, since these are not used in our app
- Rename CSS variables to `--lp-component-*` syntax that's used elsewhere

### DIFF
--- a/packages/toggle/__tests__/Toggle.cy.tsx
+++ b/packages/toggle/__tests__/Toggle.cy.tsx
@@ -12,32 +12,29 @@ describe('Toggle', () => {
   it('is accessible', () => {
     cy.mount(
       <>
-        <label htmlFor="toggle">label</label>
-        <Toggle id="toggle" />
+        <Toggle aria-label="toggle" />
       </>
     );
     cy.checkA11y();
   });
 
   it('can be reached with the keyboard', () => {
-    cy.mount(<Toggle />);
+    cy.mount(<Toggle defaultSelected={false} />);
 
     cy.getByTestId(TOGGLE).focus();
     cy.getByTestId(TOGGLE).type(' ');
-    cy.getByTestId(TOGGLE).should('be.checked');
   });
 
   it('renders an unchecked Toggle', () => {
-    cy.mount(<Toggle />);
+    cy.mount(<Toggle isSelected={false} />);
 
     cy.getByTestId(TOGGLE).should('not.be.checked');
     cy.getByTestId(TOGGLE).should('not.have.attr', 'checked', '');
   });
 
   it('renders a checked Toggle', () => {
-    cy.mount(<Toggle />);
+    cy.mount(<Toggle isSelected />);
 
-    cy.getByTestId(TOGGLE).click({ force: true });
     cy.getByTestId(TOGGLE).should('be.checked');
   });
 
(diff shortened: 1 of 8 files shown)


packages/toggle/__tests__/Toggle.cy.tsx
```diff
@@ -12,32 +12,29 @@ describe('Toggle', () => {
...
     cy.checkA11y();
   });
 
   it('can be reached with the keyboard', () => {
-    cy.mount(<Toggle />);
+    cy.mount(<Toggle defaultSelected={false} />);
 
     cy.getByTestId(TOGGLE).focus();
     cy.getByTestId(TOGGLE).type(' ');
-    cy.getByTestId(TOGGLE).should('be.checked');
```
vroske-ld, line 27: there should probably still be an assertion in here, is `not.be.checked` what you're after?

packages/toggle/src/Toggle.tsx
```diff
@@ -61,63 +52,41 @@ type ToggleProps = {
...
-  const { inputProps } = useSwitch(props, state, inputRef);
+  const { inputProps } = useSwitch({ ...props, id }, state, inputRef);
+
   const classes = cx(
     styles.Toggle,
     className,
-    checked && styles['Toggle--on'],
-    disabled && styles['Toggle--disabled']
+    state.isSelected && styles['Toggle--on'],
+    isDisabled && styles['Toggle--disabled']
```
vroske-ld, line 66: i know i ran into this with the select components, that react-aria wants to use `isDisabled` under the hood instead of `disabled`. besides that `Restrict` component in gonfalon, i'm not sure if there are other scenarios where we'd run into a mismatch of expectations between `disabled`/`isDisabled`. i wonder if it's worth us always exposing `disabled` and mapping to `isDisabled` under the hood for react-aria, or if we should use `isDisabled` for launchpad components until the need arises to switch to `disabled`?

packages/toggle/src/Toggle.tsx
```diff
@@ -61,63 +52,41 @@ type ToggleProps = {
...
   );
 
-  const handleChange = () => {
-    if (disabled || !onChange) {
-      return;
-    }
-
-    onChange();
-  };
+  console.log(inputProps);
```
vroske-ld, line 69: did you mean to leave this here?

packages/toggle/src/Toggle.tsx
chasedarkly: I believe it makes more sense to keep it as `isDisabled` and handle mapping in consuming apps if needed. Under the hood, the property is doing more business logic work the native prop `disabled` might imply. Also, other boolean "flag" props use the `is` prefix internally. Not everywhere, but I'd like to continue moving toward that pattern where it makes sense.

packages/toggle/src/Toggle.tsx
chasedarkly: nope :D

packages/toggle/src/Toggle.tsx
vroske-ld: that's reasonable. if using `isDisabled` is the preference, then in the near future we probably want to update `Restrict` to handle that; i had avoided doing that so as not to impact a bunch of components in gonfalon, but maybe it's the right move now.

chasedarkly:
ptal @vroske-ld 😄

vroske-ld (approved):

PR #802, opened 2023-03-29
fix(dropdown): merge menu trigger refs

pheggeseth:
## Summary

Supplying a `ref` to the button component used as the menu trigger/target in `Dropdown` was not working, as `Dropdown` was, internally, overwriting the ref with its own `triggerRef`. This uses the simple `mergeRefs` utility from `react-aria` to merge the local ref with the user-supplied one, if any.

## Testing approaches
Added a unit test for this.


chasedarkly (approved):
looks good, just need to pull in latest from main and resolve conflicts in the lockfile.

packages/dropdown/src/Dropdown.tsx
```diff
@@ -92,7 +94,9 @@ const Dropdown = <T extends string | object | number>(props: DropdownProps<T>) =
   const parseChildren = () => {
     const [targetChild, contentChild] = Children.toArray(children);
     return {
-      target: targetChild as ReactElement,
+      target: targetChild as FunctionComponentElement<
+        AriaAttributes & { ref: ForwardedRef<HTMLElement | undefined>; isopen: string }
+      >,
```
hisuida, line 99: Can you explain what the `isopen` is? It's a bit confusing since we have `isOpen` as well. Thanks!

pheggeseth: It's a prop for the underlying `PopoverTarget`, I believe.

hisuida: Ah, I see. Thanks!

hisuida: Is it supposed to be a boolean? (looking at `Popover.tsx, line 86`

chasedarkly: yea, this looks strange to me too. Why was this needed?

pheggeseth: The prop was being passed to the element clone anyway. `ReactElement` doesn't include a `ref` property, so I needed to cast `target` to something else, and based on what I saw in the react types, `FunctionComponentElement` seemed to fit. `ReactElement` has `any` as the default value of its props generic, but `FunctionComponentElement` has no default props generic value, so I needed to specify that `isopen` was a prop we expect, along with `AriaAttributes`, because we provide those props to the element when we clone it.

pheggeseth: > Is it supposed to be a boolean? (looking at `Popover.tsx, line 86`

No idea. I'm guessing that bit is very old code that's been preserved into launchpad. but yeah, it's actually supplying a `string` instead, and that seemed weird to me, but I didn't want to mess with things that didn't need messing. Other than the `mergeRefs` thing here, I've only altered typings to fit.

PR #809, opened 2023-04-03
fix(popover): merge target refs

pheggeseth:
## Summary

The `Popover` component was not merging user-supplied `ref`s with its own internal ref for the popover target element. That means that attaching a ref to the element/component supplied as the popover target would not take effect. Updating this is necessary for our a11y audit, as sometimes we need to manually refocus a popover target from the outside.

## Testing approaches

Added a unit test for this case.


hisuida (approved):

goyanespaula (approved):
👍

PR #810, opened 2023-04-03
feat(dropdown): allow disabling focus restoration behavior with a new "restoreFocus" prop

pheggeseth:
## Summary

There is one case, with our `SavedDashboards` feature, where we trigger the rendering of a dropdown both by pressing on the "main button" of a split button, as well as the "secondary button".

Screenshot 2023-04-03 at 2 17 12 PM

For accessibility, we need to refocus the MAIN button, and not the secondary button, if that is the button we clicked on the make the dropdown appear. The `SplitButton` component, which uses `Dropdown` internally, automatically returns focus to the _secondary_ button, because that is supposed to be the "trigger" element for the dropdown.

Rather than spend a lot of extra time refactoring a rarely-used feature, this PR introduces a simple `restoreFocus` prop to `Dropdown` which, if `false` (default is `true`), disables the normal focus restoration behavior, so we can manually focus on the main button (or somewhere else), if necessary.

## Testing approaches

Added a unit test for this behavior.

### DIFF
--- a/.changeset/brown-moons-press.md
+++ b/.changeset/brown-moons-press.md
@@ -1,0 +1,6 @@
+---
+'@launchpad-ui/dropdown': patch
+'@launchpad-ui/core': patch
+---
+
+[Dropdown]: allow disabling focus restoration behavior via the "restoreFocus" prop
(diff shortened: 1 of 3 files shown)


hisuida (approved):

pheggeseth:
After talking to @chasedarkly, we realized we could use `ButtonGroup` instead of `SplitButton` for this feature, which means we don't need this one-off special prop. I'm going to going to close this PR then.

PR #815, opened 2023-04-11
fix(popover): remove ref merging logic

chasedarkly:
Reverts a change made here: https://github.com/launchdarkly/launchpad-ui/pull/809


pheggeseth (approved):
Thanks. FYI this is a quick fix to an issue where wrapping component in `Popover` that do not forward their `ref`s was causing React Router to throw errors warning about ref forwarding, breaking unit tests.

PR #820, opened 2023-04-12
feat(clipboard): add experimental variants "basic" and "minimal"

jagarnica:
Basic Variant: 
Screenshot 2023-04-12 at 3 42 57 PM
Minimal Variant:
Screenshot 2023-04-12 at 3 43 06 PM

This adds some experimental variants for the "CopyToClipboard" component. To enable the new styles, set the "variant" prop to "basic" or "minimal". To keep using the previous version, leave the prop blank.


packages/clipboard/src/CopyCodeButton.tsx
```diff
@@ -8,19 +10,27 @@ import styles from './styles/CopyCodeButton.module.css';
 type CopyCodeButtonProps = ButtonHTMLAttributes<HTMLButtonElement> & {
   children: ReactNode;
   'data-test-id'?: string;
+  variant?: CopyToClipboardProps['variant'];
```
chasedarkly, line 13: Could we change this name to `kind` to be consistent with other components, and then add `default` to the kind as well since it's technically just another kind?

packages/clipboard/src/CopyCodeButton.tsx
```diff
@@ -8,19 +10,27 @@ import styles from './styles/CopyCodeButton.module.css';
...
       <button
         ref={ref}
         data-test-id={testId}
         type="button"
-        className={cx(styles['CopyCodeButton'], className)}
+        className={cx(styles['CopyCodeButton'], className, variant && styles[variant])}
         {...rest}
       >
         {children}
+        {variant && (
```
chasedarkly, line 27: As a second part of the first suggestion, can we then check here if kind === 'basic' || kind === 'minimal'? I like if we can make this a little more explicit.

chasedarkly (changes requested):

chasedarkly:
I think this icon needs to be translated down by 1 or 2 px
231605535-f6481c90-c9a5-4246-8cef-fd658883abaa

chasedarkly (approved):

PR #828, opened 2023-04-14
feat(filter): add disabled prop to disable Filter

jagarnica:
## Summary

This adds the ability to mark the button as disabled when appropriate and prevent the dropdown from opening when it is so. Currently, the filter button cannot be recognized as inactive by screen readers but this remedies that issue.

### DIFF
--- a/packages/filter/src/Filter.tsx
+++ b/packages/filter/src/Filter.tsx
@@ -28,6 +28,7 @@ type FilterProps = Pick<MenuProps<string>, 'size' | 'enableVirtualization'> & {
   onSelect?(item: FilterOption): void;
   isEmpty?: boolean;
   isLoading?: boolean;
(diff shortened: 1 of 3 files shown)


packages/filter/src/Filter.tsx
```diff
@@ -67,12 +69,18 @@ const Filter = ({
   };
 
   return (
-    <Dropdown targetTestId={testId} targetClassName={dropdownClasses} {...props}>
+    <Dropdown
+      targetTestId={testId}
+      disabled={disabled}
+      targetClassName={dropdownClasses}
+      {...props}
+    >
```
jagarnica, line 77: The dropdown can no longer be opened when the "disabled" prop is sent to `Filter` component.

chasedarkly (approved):

PR #837, opened 2023-04-20
fix: support modal + drawer combo; fix drawer focus trap

cjhensen:
## Summary

**Drawers**
The focus trap in all drawers is currently broken - you can tab out into the browser.
The fix for this was to render the portal separately from the drawer, in the same fashion as the modal component. The cause is still not 100% clear, but is likely due to some combination of context or re-rendering issues.

**Drawer + Modal combo** 
When a modal is triggered from a button inside a drawer, the event handlers become funky as both components are listening for the same Escape key event.
The fix: add/remove the event listeners based on if focus is within the drawer or modal.

**Screen reader accessibility for drawers and modals**
Changed aria-labelledby to aria-describedby so screen readers can properly announce the modal title

### DIFF
--- a/packages/drawer/src/Drawer.tsx
+++ b/packages/drawer/src/Drawer.tsx
@@ -6,10 +6,11 @@ import { FocusTrap } from '@launchpad-ui/focus-trap';
 import { Close } from '@launchpad-ui/icons';
 import { Portal } from '@launchpad-ui/portal';
 import { Progress } from '@launchpad-ui/progress';
+import { useFocusWithin } from '@react-aria/interactions';
 import { usePreventScroll } from '@react-aria/overlays';
 import { cx } from 'classix';
 import { LazyMotion, m } from 'framer-motion';
-import { Suspense, useEffect, useRef } from 'react';
+import { useState, Suspense, useEffect, useRef } from 'react';
 
 import { DRAWER_LABELLED_BY } from './constants';
 import styles from './styles/Drawer.module.css';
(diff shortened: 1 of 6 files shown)


packages/drawer/src/Drawer.tsx
```diff
@@ -45,7 +46,13 @@ type DrawerProps = {
   hideCancel?: boolean;
 };
 
-const Drawer = ({
+const Drawer = (props: DrawerProps) => (
+  <Portal>
+    <DrawerContainer {...props} />
+  </Portal>
+);
```
cjhensen, line 53: rendering the portal separately from the drawer contents fixed the issue of the broken focus trap

packages/drawer/src/Drawer.tsx
```diff
@@ -55,30 +62,48 @@ const Drawer = ({
...
+      removeOverlayAndEventHandler();
+    }
 
     return () => {
-      document.body.classList.remove('has-overlay');
-      document.removeEventListener('keydown', handleEscape);
+      removeOverlayAndEventHandler();
     };
-  }, [onCancel, testId]);
+  }, [onCancel, testId, isFocusWithin]);
```
cjhensen, line 106: this is what allows for smoother drawer + modal focus behavior. adding/removing event listeners based on if the element is focused within

sunnyguduru (approved):
I don't have much context into drawers but the modal code looks good

apucacao (approved):
This make sense.

I wonder if Modal and Drawer should share some of their "modality" in our implementation: they're so similar.

PR #845, opened 2023-04-24
Revert "feat(menu): deprecate menuitemlink in favor of asChild"

kwatkins-ld:
Reverts launchdarkly/launchpad-ui#835.

The original PR is causing unit test failures in consumers, after spending some time investigating I think we should revert so that we're not blocked releasing new work while doing analysis on the problem(s) this work caused.


stinachen (approved):

kwatkins-ld:
@jagarnica was able to fix this forward in #846 , no need to revert

PR #846, opened 2023-04-24
fix(menu): Allow different size shaped menu items when virtual

jagarnica:
## Summary

Items were getting clipped when virtualized such as when adding a profile picture or other element in the menu item. This will allows to have dynamic items.

### DIFF
--- a/packages/menu/src/Menu.tsx
+++ b/packages/menu/src/Menu.tsx
@@ -192,7 +192,7 @@ const ItemVirtualizer = <T extends number | string>(props: ItemVirtualizerProps<
   const {
     overscan,
     searchElement,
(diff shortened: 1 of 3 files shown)


packages/menu/src/Menu.tsx
```diff
@@ -192,7 +192,7 @@ const ItemVirtualizer = <T extends number | string>(props: ItemVirtualizerProps<
   const {
     overscan,
     searchElement,
-    itemHeight = 33,
+    itemHeight = 31.5,
```
jagarnica, line 195: This is the actual measured height for a menu item when used with just text.

bjlagman12 (approved):

PR #851, opened 2023-05-10
feat: use ComponentProps type to simplify intrinsic element props

kwatkins-ld:
## Summary

We can replace the use of `ElementAttributesType<ElementType>` by using the built-in `ComponentProps` utility type in React.

I found that this doesn't work well with prop types that get passed to `forwardRef`, it might be worth addressing that in a future pass but for now this is a simple change that I think reduces overall complexity.

### DIFF
--- a/packages/avatar/src/Avatar.tsx
+++ b/packages/avatar/src/Avatar.tsx
@@ -1,5 +1,5 @@
 import type { IconProps } from '@launchpad-ui/icons';
-import type { ComponentType, HTMLAttributes } from 'react';
+import type { ComponentType, ComponentProps } from 'react';
 
 import { Person } from '@launchpad-ui/icons';
 import { cx } from 'classix';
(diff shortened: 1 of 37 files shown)


packages/avatar/src/Avatar.tsx
```diff
@@ -8,7 +8,7 @@ import { useCallback, useEffect, useState } from 'react';
 import styles from './styles/Avatar.module.css';
 import { useIsMounted } from './utils';
 
-type AvatarProps = HTMLAttributes<HTMLDivElement> & {
+type AvatarProps = ComponentProps<'img'> & {
```
kwatkins-ld, line 11: This was previously typed as possible `div` attributes, but we only render a `div` if the `url` prop is omitted (or, if we attempt to load the image at the given `url` and get back a 404 or a format we don't support).

Changing the prop type to use `ComponentProps<'div'>` uncovered this, and it seems safer to just accept the intrinsic props for `img` and forward them along (knowing that some like `srcset` won't have any effect if we end up rendering a `div`).

apucacao (approved):
🎉

portpaw (approved):

Niznikr:
👍 @kwatkins-ld for a follow-up, seems `ComponentPropsWithRef` might work well for `forwardRef` [as described here](https://github.com/typescript-cheatsheets/react/blob/main/docs/advanced/patterns_by_usecase.md#componentprops)

PR #853, opened 2023-05-12
feat(form): add `useNumberField` hook

pheggeseth:
## Summary

This adds a new `useNumberField` hook to the `form` package, which can be used to instantiate a number input component using `@react-aria`'s number field hooks.

Why a hook instead of a component? Because, in order to wire the relevant elements together using the correct aria attributes, we need to be able to spread props for the number field's corresponding label, form him (description), or field error on their corresponding elements. Directly rendering a `NumberField` component would make it impossible to do this.

So instead, this hook can be used like this:
```jsx
const { labelProps, renderNumberField } = useNumberField();

return (
  <>
    <Label {...labelProps}>Number field</Label>
    {renderNumberField()}
  </>
); 
```

or, given the changes made to the `FormField` component here, we can do:
```jsx
const [value, setValue] = useState(123);

const { 
  labelProps, 
  formHintProps, 
  fieldErrorProps, 
  renderNumberField,
} = useNumberField({ id: 'number-field', value, onChange: setValue });

return (
  <FormField
    label="Number field"
    name="Number field"
    htmlFor="number-field"
    LabelProps={...labelProps}
    FormHintProps={...formHintProps}
    FieldErrorProps={...fieldErrorProps}
  >
    {renderNumberField()}
  </FormField>
); 
```


pheggeseth:
Yeaaaa, it lives! 🎉

Niznikr (approved):
Looks like a solid approach and I do appreciate the versatility it provides. Curious what folks think of the DX but we can always tweak things if needed 👍

pheggeseth:
> Looks like a solid approach and I do appreciate the versatility it provides. Curious what folks think of the DX but we can always tweak things if needed 👍

Just some background that part of my thinking is that we don't currently have a single pattern for using a text-based form field. Sometimes we use a bare `TextField` pared with a `Label`, and in some places we use our `FormField` component to include labels and descriptions for an input. These hooks from React Aria are designed with the expectation that the consumer will have access to things like `Label` and `FormHint` internal components. I didn't feel it was worth it to reinvent the `FormField` wheel and create a `NumberField` component at this time, because while we want to have access to the input itself (provided here by a render prop), in some cases we will only be using some of the other props objects returned by this hook (such as the `labelProps`), and in other cases we would use all of them (like for forms that already use the `FormField` component). Providing this as a hook leaves the decision of how much of the API to use up to the consumer, increasing effort, but also increasing flexibility.

With that in mind, this was also originally intended to be a relatively short-term solution for fixing a long-standing bug in our platform regarding flag variation types and how we represent number in inputs in the UI. We had an incident recently as a result of a very vocally unhappy customer, and I understand that we get support escalations regarding this edge case somewhat regularly.

AFAIC, we could take this work and create a simple `NumberField` wrapper within `gonfalon` for using this hook with `FormField`, if it was needed.

PR #864, opened 2023-06-05
refactor(core)!: use named exports

Niznikr:
## Summary

To improve tree-shaking and support [Remix's CSS bundling](https://remix.run/docs/en/main/guides/styling#css-bundling), replace barrel exports with name exports in the package's index file. Remove the icons and tokens packages from `core` as their exports are auto-generated and would incur too high of maintenance cost to `core`'s index file. This is [commonly done](https://github.com/primer/react/blob/main/src/index.ts) in other [bundled packages](https://github.com/adobe/react-spectrum/blob/main/packages/react-aria/src/index.ts) in the community.

## Testing approaches

CI passes


apps/remix/app/root.tsx
```diff
@@ -15,73 +16,15 @@ import {
...
-import sliderStyles from '@launchpad-ui/slider/style.css';
-import snackbarStyles from '@launchpad-ui/snackbar/style.css';
-import splitButtonStyles from '@launchpad-ui/split-button/style.css';
-import tabListStyles from '@launchpad-ui/tab-list/style.css';
-import tableStyles from '@launchpad-ui/table/style.css';
-import tagStyles from '@launchpad-ui/tag/style.css';
-import toastStyles from '@launchpad-ui/toast/style.css';
-import toggleStyles from '@launchpad-ui/toggle/style.css';
-import tooltipStyles from '@launchpad-ui/tooltip/style.css';
-// plop end imports
```
Niznikr, line 49: Huge DX win for Remix integration!

erangeles: @Niznikr I was thinking of proposing this in gonfalon, to add an eslint rule to default to named exports, any thoughts on this?

erangeles (approved):
🔥

apps/remix/app/root.tsx
erangeles: that's what I was thinking, on all new additions as to not be a pain, or manually add an ignore rule statement on the top so we can remember to peel them off incrementally! thanks for your advice!

kwatkins-ld: 👋 In general I'm a fan of using named exports anywhere we can... to y'alls point(s) above we can add exclusions for valid exceptions as necessary if we choose to go this route.

Sorta related, we've used code-mods a couple of times now to bulk-exclude existing violations of a new lint rule while also turning it on, to be able to prevent the spread while also reducing dev friction ([example PR](https://github.com/launchdarkly/gonfalon/pull/27440/files).

.eslintrc.js
```diff
@@ -78,6 +78,10 @@ module.exports = {
         selector: 'TSQualifiedName[left.name=React]',
         message: "Don't use the React namespace. Import the type instead.",
       },
+      {
+        selector: ':matches(ExportAllDeclaration)',
```
Niznikr, line 82: Rule to error on `*` exports.

apps/remix/app/root.tsx
erangeles: thanks for the example P.R, this is very useful to know. I've document this effort in a shortcut [ticket](https://app.shortcut.com/launchdarkly/story/205508/create-eslint-rule-to-enforce-named-exports)

PR #872, opened 2023-06-07
perf: enable storyStoreV7 to lazy load stories

Niznikr:
## Summary

Enable `storyStoreV7` to lazy load stories which speeds up local dev and our built Storybook. To avoid issues with [code splitting CSS modules](https://github.com/vitejs/vite/issues/3924) disable it for builds.

## Testing approaches

Chromatic passes


apucacao (approved):

Niznikr:
Looks like we still get ordering issues based on the Chromatic results

PR #880, opened 2023-06-14
feat: add stack and types packages

Niznikr:
## Summary

Add `types` package to share common types between components. Add `stack` package to space components.


packages/stack/src/styles/Stack.module.css
```diff
@@ -0,0 +1,60 @@
...
+
+.center {
+  align-items: center;
+}
+
+.right {
+  align-items: flex-end;
+}
+
+.gap-0 {
```
apucacao, line 26: I know this is the first pass, but while we're at it, what do you think about switching to `gap` instead of padding? That's really what we want here.

apucacao (approved):

PR #921, opened 2023-07-12
fix(textField): disable password managers from adding autofill icon to input

hisuida:
## Summary
- If `autoComplete = "off"` for `TextField` input, 1Password will not inject autofill options and icons to the field
- I tried to implement this for LastPass as well, but recent updates disabled the fix, so as of now, there is no option to disable LastPass :(

Without fix:
Screenshot 2023-07-12 at 3 00 06 PM

With Password manager disabled:
Screenshot 2023-07-12 at 3 00 12 PM

## Testing approaches
- manual

### DIFF
--- a/packages/form/src/TextField.tsx
+++ b/packages/form/src/TextField.tsx
@@ -24,6 +24,7 @@ const TextField = forwardRef<HTMLInputElement, TextFieldProps>(
       suffix,
       overrideWidth,
       'data-test-id': testId = 'text-field',
+      autoComplete,
       ...rest
     },
     ref
@@ -39,6 +40,7 @@ const TextField = forwardRef<HTMLInputElement, TextFieldProps>(
             {...rest}
             type={type}
             data-test-id={testId}
+            autoComplete={autoComplete}
             className={classes}
             readOnly={readOnly}
             ref={ref}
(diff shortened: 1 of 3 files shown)


packages/form/src/TextField.tsx
```diff
@@ -51,13 +53,39 @@ const TextField = forwardRef<HTMLInputElement, TextFieldProps>(
       );
     }
 
+    if (autoComplete === 'off') {
+      return (
+        <input
+          {...rest}
+          data-lpignore="true" // "data-lpignore" is added to prevent LastPass from injecting a password autofill icon
+          data-1p-ignore // "data-1p-ignore" is added to prevent 1Password from injecting a password autofill icon
```
Niznikr, line 61: Could we instead conditionally apply these attributes based off `autoComplete` to the template below to minimize duplication?

hisuida: Yeah! I'll update :)

Niznikr (approved):

hisuida:
Looks like with LastPass' newest update, the `data-lpignore` is disabled. It seems like adding `search` to the IDs of inputs may work but as of now, I'll go ahead and merge the fix for 1Password in and see if there are any other definite fixes available for LastPass

PR #930, opened 2023-07-20
feat(icons): add unique id prefix and allow empty path

Niznikr:
## Summary

For consumers hosting static assets in a cdn or different domain we should allow them to inject the sprite into the document and have `Icon` access it there. Avoid id collisions by adding a `lp-icon` prefix to all symbols.

### DIFF
--- a/.changeset/cuddly-cows-applaud.md
+++ b/.changeset/cuddly-cows-applaud.md
@@ -1,0 +1,5 @@
+---
+'@launchpad-ui/icons': patch
+---
+
+Add unique id prefix and allow empty path
--- a/packages/icons/src/Icon.tsx
+++ b/packages/icons/src/Icon.tsx
@@ -53,7 +53,9 @@ const Icon = ({
       >
         {title && <title id={titleId}>{title}</title>}
         {description && <desc id={descriptionId}>{description}</desc>}
-        <use href={`${spritePath || '/static/sprite.svg'}#${name}`} />
+        <use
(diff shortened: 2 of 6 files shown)


packages/icons/src/Icon.tsx
```diff
@@ -38,7 +38,10 @@ const Icon = ({
   const isAriaHidden = ariaHidden ?? (!ariaLabelledBy && !ariaLabel);
   const titleId = title && `${prefix}-${name}-title`;
   const descriptionId = description && `${prefix}-${name}-description`;
-  const { path: spritePath } = useContext(IconContext);
+
+  const { path: contextPath } = useContext(IconContext);
+  const iconId = `lp-icon-${name}`;
```
portpaw, line 43: Could this ID be part of the context value, and then either provide a path or an ID to that context?

Niznikr: True we can have it default to `lp-icon` but open it up to custom ones too 👍

portpaw: Oh sorry, I actually had a fundamental misunderstanding of what this ID was doing. I thought this was the ID for the sprite SVG, not the individual icon within that sprite.

I guess when it's injected into the DOM, we don't need to know the sprite's ID, since each icon has its own ID that we can reference globally.

portpaw (approved):

.changeset/cuddly-cows-applaud.md
portpaw, line 1: Cuddly cows applaud ☺️  🐮 👏

apucacao (approved):

PR #938, opened 2023-07-24
feat: add inline-edit package

Niznikr:
## Summary

Add `inline-edit` package to display and allow inline editing of a form elements:

- Use props `defaultValue` and `onConfirm` to handle state management of the value to edit
- Have children act as the "read" view of the component
- Hide edit icon button and wrap children with a React Aria button when `hideEdit` is true
- Implement focus management to ensure focus is directed correctly when toggling between read and edit mode
- Use `renderInput` prop to allow passing a custom `TextField` or `TextArea` component
- Add handlers for edit, cancel, and confirm actions
- Use prop `isEditing` to allow full control over the read and edit modes
- Add `@vanilla-extract/css` as a peer dependency for prop `layout` variant types
- Use `useFocusWithin` to cancel edit on blur

## Testing approaches

https://626696a2018c1f004a1cde86-biuafqdqpl.chromatic.com/

### DIFF
--- a/.changeset/chatty-tables-attend.md
+++ b/.changeset/chatty-tables-attend.md
@@ -13,3 +13,4 @@ Add `inline-edit` package to display and allow inline editing of a form elements
 - Add handlers for edit, cancel, and confirm actions
 - Use prop `isEditing` to allow full control over the read and edit modes
 - Add `@vanilla-extract/css` as a peer dependency for prop `layout` variant types
+- Use `useFocusWithin` to cancel edit on blur
--- a/.changeset/chatty-tables-attend.md
+++ b/.changeset/chatty-tables-attend.md
@@ -1,0 +1,12 @@
+---
+'@launchpad-ui/inline-edit': minor
+'@launchpad-ui/core': patch
+---
+
+Add `inline-edit` package to display and allow inline editing of a form elements:
+
+- Use props `defaultValue` and `onSave` to handle state management of the value to edit
+- Have children act as the "read" view of the component
+- Hide edit icon button and wrap children with a React Aria button when `hideEdit` is true
+- Implement focus management to ensure focus is directed correctly when toggling between read and edit mode
+- Use `input` prop to allow passing a custom `TextField` or `TextArea` component
--- a/.changeset/chatty-tables-attend.md
+++ b/.changeset/chatty-tables-attend.md
@@ -9,4 +9,4 @@ Add `inline-edit` package to display and allow inline editing of a form elements
 - Have children act as the "read" view of the component
 - Hide edit icon button and wrap children with a React Aria button when `hideEdit` is true
 - Implement focus management to ensure focus is directed correctly when toggling between read and edit mode
-- Use `input` prop to allow passing a custom `TextField` or `TextArea` component
+- Use `renderInput` prop to allow passing a custom `TextField` or `TextArea` component
--- a/.changeset/chatty-tables-attend.md
+++ b/.changeset/chatty-tables-attend.md
@@ -5,8 +5,11 @@
 
 Add `inline-edit` package to display and allow inline editing of a form elements:
 
-- Use props `defaultValue` and `onSave` to handle state management of the value to edit
+- Use props `defaultValue` and `onConfirm` to handle state management of the value to edit
 - Have children act as the "read" view of the component
 - Hide edit icon button and wrap children with a React Aria button when `hideEdit` is true
 - Implement focus management to ensure focus is directed correctly when toggling between read and edit mode
 - Use `renderInput` prop to allow passing a custom `TextField` or `TextArea` component
+- Add handlers for edit, cancel, and confirm actions
+- Use prop `isEditing` to allow full control over the read and edit modes
+- Add `@vanilla-extract/css` as a peer dependency for prop `layout` variant types
--- a/packages/inline-edit/src/InlineEdit.tsx
+++ b/packages/inline-edit/src/InlineEdit.tsx
@@ -1,6 +1,6 @@
 import type { InlineVariants } from './styles/InlineEdit.css';
 import type { ButtonProps } from '@launchpad-ui/button';
-import type { ComponentProps, Dispatch, SetStateAction } from 'react';
+import type { ComponentProps, Dispatch, KeyboardEventHandler, SetStateAction } from 'react';
 
 import { Button, ButtonGroup, IconButton } from '@launchpad-ui/button';
 import { TextField } from '@launchpad-ui/form';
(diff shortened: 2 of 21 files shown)


packages/inline-edit/src/styles/InlineEdit.css.ts
```diff
@@ -0,0 +1,50 @@
...
+  },
+  ':focus-visible': {
+    borderRadius: vars.border.radius.medium,
+    boxShadow: `0 0 0 2px ${vars.color.bg.ui.primary}, 0 0 0 4px ${vars.color.shadow.interactive.focus}`,
+    outline: 0,
+    zIndex: 2,
+  },
+});
+
+type InlineVariants = RecipeVariants<typeof inline>;
```
Niznikr, line 47: This Vanilla Extract utility returns type `layout: "vertical" | "horizontal"` for the prop which means we can have a single source of truth for props tied directly to component styles (variants).

.changeset/chatty-tables-attend.md
```diff
@@ -0,0 +1,12 @@
+---
+'@launchpad-ui/inline-edit': minor
+'@launchpad-ui/core': patch
+---
+
+Add `inline-edit` package to display and allow inline editing of a form elements:
+
+- Use props `defaultValue` and `onSave` to handle state management of the value to edit
```
apucacao, line 8: This means this component cannot be "controlled", right?

Are there use  cases where we might want it to be controlled too?

.changeset/chatty-tables-attend.md
```diff
@@ -0,0 +1,12 @@
...
+- Have children act as the "read" view of the component
+- Hide edit icon button and wrap children with a React Aria button when `hideEdit` is true
+- Implement focus management to ensure focus is directed correctly when toggling between read and edit mode
+- Use `input` prop to allow passing a custom `TextField` or `TextArea` component
```
apucacao, line 12: Would naming this `renderInput` make this self-explanatory?

packages/inline-edit/src/InlineEdit.tsx
```diff
@@ -0,0 +1,137 @@
...
+  'data-test-id': testId = 'inline-edit',
+  layout = 'horizontal',
+  children,
+  defaultValue,
+  onSave,
+  hideEdit = false,
+  input = <TextField />,
+  'aria-label': ariaLabel,
+}: InlineEditProps) => {
+  const [isEditing, setEditing] = useState(false);
```
apucacao, line 41: Same question the one about `defaultValue` and un/controlled.

Are there use cases where it would be useful to pass `isEditing` from the owner component?

apucacao (commented):
Working well so far!

.changeset/chatty-tables-attend.md
Niznikr: Yeah having controlled can be useful! Will add it.

packages/inline-edit/stories/InlineEdit.stories.tsx
```diff
@@ -0,0 +1,137 @@
...
+        isEditing={isEditing}
+        onCancel={() => setEditing(false)}
+        onEdit={() => setEditing(true)}
+        {...args}
+        onConfirm={(value) => {
+          setEditValue(value);
+          setEditing(false);
+        }}
+      >
+        <span>{editValue}</span>
```
apucacao, line 133: Did you get a chance to try using a render function for `children`? The idea was that in uncontrolled mode, it would remove the need for consumers to manage the state of the field's value.

packages/inline-edit/src/InlineEdit.tsx
```diff
@@ -0,0 +1,158 @@
...
+    !controlled && setEditing(false);
+    onCancel?.();
+  };
+
+  const handleConfirm = () => {
+    onConfirm(inputRef.current?.value || '');
+    !controlled && setEditing(false);
+  };
+
+  const handleKeyDown: KeyboardEventHandler<HTMLInputElement> = (event) => {
```
kwatkins-ld, line 85: Was exploring the new stories in this branch locally, noticed that pressing `Esc` will cancel an in-progress edit which brought me here, wondering if we should also consider the case where a user blurs away from the input/cancel/save buttons while editing (either by clicking outside of the affected elements, or via keyboard navigation)?

There is some nuance to consider, but I'm wondering about the case where multiple inline-editable fields exist next to each other... how would a user navigate between them? For example: if I begin editing a field but then try to keyboard navigate to another editable field, what should happen? I'd expect the edit for the first field to be implicitly cancelled, but curious to hear your thoughts.

Niznikr: Yeah no strong opinion on blur implicitly cancelling. Can try that out and see how that feels.

packages/inline-edit/stories/InlineEdit.stories.tsx
Niznikr: I did but was struggling with the DX. Do you have an idea what something like that would look like? Seems like folks would still need to use `onConfirm` to have access to the updated value to do something with it form/data wise.

packages/inline-edit/src/InlineEdit.tsx
kwatkins-ld: > How does this overlap with routing?

Today we consider a form dirty if the input hasn't been saved, and dirty forms (in most places) cause route changes to be blocked and the user is forced to confirm the navigation... if we make this interaction modal, blurring a field that is being edited would discard the draft (and therefore not cause the field to be considered dirty).

Conversely, if this interaction isn't modal the following scenario seems possible: begin editing a field, change the field's value, then navigate to a different field... in that case the first field would need to be considered dirty to prevent unexpectedly losing edits.

I also agree that these questions aren't especially complicated, and that we don't necessarily have to answer them in this PR.

apucacao: That said, we should be opinionated here before we show up at UX crit.

My 2¢: Let's cancel on blur for now. That feels snappier to me. 😅

apucacao:
Any thoughts in making the entire component interactive when in non-editable mode? Clicking could enable editable mode. That would give us a larger click area, and bring us closer to the native contenteditable behavior. We could also give some subtle treatment to that interaction to make that clear.

apucacao (approved):
This looks good to me.

There's lot of potential tweaks we can make to the interaction, but we can do that once this is already in people's hands.

packages/inline-edit/src/InlineEdit.tsx
Niznikr: Meeting with Antonio shortly and can go over this topic.

Niznikr: Discussed with Antonio and we are gonna go with cancel on blur 👍

lwonsower (approved):
This is awesome. 🙏🏼

PR #946, opened 2023-08-02
feat(inline-edit): forward ref and spread additional props

Niznikr:
## Summary

Forward the ref of the edit input and spread additional props onto the container.

### DIFF
--- a/packages/inline-edit/src/InlineEdit.tsx
+++ b/packages/inline-edit/src/InlineEdit.tsx
@@ -1,12 +1,6 @@
 import type { InlineVariants } from './styles/InlineEdit.css';
 import type { TextAreaProps, TextFieldProps } from '@launchpad-ui/form';
-import type {
-  ComponentProps,
-  Dispatch,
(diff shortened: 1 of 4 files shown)


packages/inline-edit/src/InlineEdit.tsx
```diff
@@ -35,133 +29,152 @@ type InlineEditProps = ComponentProps<'div'> &
...
+        ref: mergeRefs(inputRef, ref),
+        defaultValue,
+        onKeyDown: handleKeyDown,
+        'aria-label': ariaLabel,
+      })
+    );
+
+    return isEditing ? (
+      <div
+        {...rest}
```
apucacao, line 142: I assume these props will always map to valid DOM attributes, right?

Niznikr: Yeah the ones coming from `ComponentProps<'div'>`.

pheggeseth (approved):

PR #966, opened 2023-08-15
Update FormField label type

XieX:
## Summary

Updates the type of the `label` prop for the `FormField` component to `React.ReactNode`.

The specific use case is including a `HelpTooltip` component after the label text, but there are other cases where you may want your label to be more than just a string.

### DIFF
--- a/packages/form/src/FormField.tsx
+++ b/packages/form/src/FormField.tsx
@@ -12,7 +12,7 @@ import styles from './styles/Form.module.css';
 
 type FormFieldProps = {
   isRequired: boolean;
-  label?: string;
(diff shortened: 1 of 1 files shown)


.changeset/slow-bikes-type.md
```diff
@@ -0,0 +1,2 @@
+---
```
Niznikr, line 1: Thanks for the contribution! Could you follow the [guidelines here](https://github.com/launchdarkly/launchpad-ui/blob/main/CONTRIBUTING.md#step-3-generate-a-changelog-for-your-code) to ensure the changelog entry properly communicates the change to consumers?

PR #971, opened 2023-08-17
ci: run a11y tests on stories

Niznikr:
## Summary

Right now we run `axe-core` on manually written Cypress component tests. Most simply test the default view/state of a component. To minimize the number of total tests we need to write and ensure we test all states (including dark mode contrast) we can run axe on our stories once Chromatic finishes a build. This would effectively replace our current Cypress component testing approach.

The workflow:

- Chromatic makes a build
- Run `axe-core` tests using Playwright that uses the Storybook's `stories.json` that lists all stories and their ids
- Shard the suite to minimize CI time (246 stories takes ~2.5 min currently in GH actions)

ToDo:

- [ ] Get [Story context](https://github.com/storybookjs/test-runner#getstorycontext) to allow configuration through individual stories
- [ ] Disable tests for stories that fail with a `@fixme` comment to address down the road
- [ ] Write ADR entry

Screenshot 2023-08-17 at 1 05 19 PM


playwright/a11y.spec.ts
```diff
@@ -0,0 +1,28 @@
...
+      const root = page.locator('#storybook-root');
+      await root.waitFor();
+
+      const accessibilityScanResults = await new AxeBuilder({ page })
+        .withTags(['wcag2a', 'wcag2aa', 'wcag21a', 'wcag21aa'])
+        .disableRules('region')
+        .include('#storybook-root')
+        .analyze();
+
+      await expect(accessibilityScanResults.violations).not.toBeNull();
```
Niznikr, line 25:
```suggestion
      await expect(accessibilityScanResults.violations).toEqual([]);
```
Update once failing stories are disabled.

vroske-ld:
Playwright returns! 🪄 😄 

i love that we can now include light + dark modes in a11y checks. (are we already doing the same for visual regression checks in Chromatic? it's been a while since i've been in this space)

a couple of questions (i imagine you'll capture some of this in your ADR):
- What makes Playwright better suited for this approach? Could we not achieve the same strategy using Cypress since both use `axe-core` under the hood?
- are there opportunities for us to run the a11y checks besides the default loaded state (i.e., after simulating an interaction)? would we still need component-level tests for that and if so which technology would we use?

Niznikr:
Yes we are already using side-by-side snapshots for visual tests 😄 Great questions! Cypress was faster when it came to [isolated component testing](https://docs.cypress.io/guides/component-testing/overview) but now since we're running e2e tests on the deployed Chromatic instance Playwright is faster and easily allows us to [shard the suite](https://playwright.dev/docs/test-sharding) to optimize things. Regarding interactions, we are using the [Storybook addon](https://storybook.js.org/docs/7.2/react/essentials/interactions) in some stories so we also get those checked by these tests. Every story and their interactions from the `play` function are covered.

stinachen (approved):

PR #993, opened 2023-09-05
feat(inline-edit): allow to render `IconField`

Niznikr:
## Summary

Allow `InlineEdit` to render `IconField`. Add `tooltip` and `renderIconLast` props to `IconField`.

Screenshot 2023-09-05 at 4 07 57 PM

### DIFF
--- a/packages/form/src/IconField.tsx
+++ b/packages/form/src/IconField.tsx
@@ -30,10 +30,9 @@ const IconField = ({
   const iconElement = cloneElement(icon, {
     size: 'small',
     className: cx(styles.iconFieldIcon, styles.iconFieldIconFill),
-    style: renderIconLast ? { right: '1rem' } : { left: '1rem' },
   });
 
-  const classes = cx(styles.iconField, renderIconLast && 'IconAfter', className);
+  const classes = cx(styles.iconField, renderIconLast ? 'IconAfter' : 'IconBefore', className);
 
   const renderIcon = tooltip ? (
     <Tooltip content={tooltip} targetClassName={styles.iconFieldButton}>
--- a/packages/form/src/IconField.tsx
+++ b/packages/form/src/IconField.tsx
@@ -1,6 +1,8 @@
 import type { IconProps } from '@launchpad-ui/icons';
 import type { ComponentProps, ReactElement } from 'react';
 
+import { IconButton } from '@launchpad-ui/button';
+import { Tooltip } from '@launchpad-ui/tooltip';
 import { cx } from 'classix';
 import { cloneElement } from 'react';
 
@@ -10,6 +12,9 @@ type IconFieldProps = ComponentProps<'div'> & {
   icon: ReactElement<IconProps>;
   children: JSX.Element | JSX.Element[];
   'data-test-id'?: string;
+  tooltip?: string | JSX.Element;
+  renderIconLast?: boolean;
+  ariaLabel?: string;
 };
 
 const IconField = ({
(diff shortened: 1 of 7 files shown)


packages/form/src/IconField.tsx
```diff
@@ -10,23 +12,49 @@ type IconFieldProps = ComponentProps<'div'> & {
...
+    size: 'small',
+    className: cx(styles.iconFieldIcon, styles.iconFieldIconFill),
+  });
 
-  const classes = cx(styles.iconField, className);
+  const classes = cx(styles.iconField, renderIconLast ? 'IconAfter' : 'IconBefore', className);
+
+  const renderIcon = tooltip ? (
+    <Tooltip content={tooltip} targetClassName={styles.iconFieldButton}>
+      <IconButton
```
apucacao, line 39: Did you mean to make this into a button? Will users interact with it? (It looks like you can't pass `onClick` to this specific component for example.)

Niznikr: Yeah tooltips must have an interactable element as the target/trigger. `onClick` isn't needed as `Tooltip` already adds listeners for hover/focus to show the content.

packages/form/src/styles/Form.module.css
```diff
@@ -88,6 +84,28 @@ select.formInput {
   border-color: var(--lp-color-border-field-error);
 }
 
+.iconField:global(.IconBefore) .formInput {
+  padding-left: 3rem;
```
apucacao, line 88: should we use the specing tokens here?

packages/inline-edit/src/InlineEdit.tsx
```diff
@@ -127,14 +127,24 @@ const InlineEdit = forwardRef<HTMLInputElement, InlineEditProps>(
...
-      mergeProps(renderInput.props, {
-        ref: mergeRefs(inputRef, ref),
-        defaultValue,
-        onKeyDown: handleKeyDown,
-        'aria-label': ariaLabel,
-      })
+      mergeProps(renderInput.props, inputChildren ? {} : inputProps),
+      inputChildren &&
+        Children.map(inputChildren, (child) =>
+          ['TextField', 'TextArea'].includes(child.type.displayName)
```
apucacao, line 144: Why do we need to reach for the display name here? Is reading `displayName` this way reliable post-bundling, etc…

pheggeseth: @apucacao, I'm guessing this is to check that, if it's an expected Launchpad-based React element, then we merge its props with the `inputProps` we need. I don't know anything about bundling effects on this.

pheggeseth (approved):
Thanks, @Niznikr! This should work for our needs. Feel free to address any other feedback from @apucacao as needed.

packages/inline-edit/src/InlineEdit.tsx
apucacao: Yeah, I figured.

I was wondering why `child.type === TextField` wouldn't work. We'd need to import `TextField`.

packages/form/src/styles/Form.module.css
Niznikr: We don't have one for 3rem 😅

packages/inline-edit/src/InlineEdit.tsx
Niznikr: LP component display names aren't minified in the product but just to be safe I can check for the type to match the component itself.

PR #1001, opened 2023-09-15
chore(icon): remove fixed size for inner icon in FlairIcon and absolu…

jennifro:
…itioning

## Summary

For `FlairIcon`, this PR removes the set sizing of `medium` for the nested inner icon and allows for any sizing of the inner `Icon`, except for `huge` or `xlarge` sizing. 

I didn't save these changes to the FlairIcon stories, I only temporarily set sizing and then reverted changes. 

Screenshot 2023-09-15 at 12 36 24 PM

Screenshot 2023-09-15 at 12 36 31 PM

## Testing approaches

Manually as well as using the testing commands in the README.md


packages/icons/src/FlairIcon.tsx
```diff
@@ -20,9 +20,19 @@ const FlairIcon = ({
   gradient = 'purpleToBlue',
   ...props
 }: FlairIconProps) => {
+  const getIconSize = () => {
+    let iconSize: IconProps['size'] = children.props.size;
+
+    if (iconSize === 'xlarge' || iconSize === 'huge' || !iconSize) {
```
Niznikr, line 26: Curios why we want to restrict any sizes in general?

jennifro: xlarge and huge go outside the confines of the flair -- I can make it so the FlairIcon is responsive to the inner icon size, would that be preferable?

Niznikr: If it isn't too much of a pain I'd say so. Would be nice to have equivalence with the standard `Icon` component.

Niznikr (approved):

PR #1030, opened 2023-10-06
refactor(button): convert to css modules

Niznikr:
## Summary

We need to tweak some variables button uses for the refresh. So to allow us to ff those styles we need to convert them to css modules.


packages/split-button/src/styles/SplitButton.module.css
```diff
@@ -5,39 +5,43 @@
...
 
+.SplitButtonDrop:global([class*='_Button_']) {
+  border-left-width: 1px;
+  border-top-left-radius: 0;
+  border-bottom-left-radius: 0;
+  padding: 0 0.8rem;
+  margin-left: -0.1rem;
+}
+
+.SplitButtonDrop:global([class*='_Button--small_']) {
```
apucacao, line 16: What are those selectors with underscores?

Niznikr: Targeting parent selectors whose class names get hashed from css modules

apucacao (approved):

PR #1031, opened 2023-10-07
feat(tokens): add primitive size tokens

matthewferry:
## Summary

Adds a new primitive for `size`. The aim of these primitives is to use them for spacing (spacing can becomes size aliases), width/heights (e.g. avatar and icon sizes should be aliases to these primitives).


packages/tokens/src/viewport.yaml
```diff
@@ -1,9 +1,9 @@
 viewport:
   mobile:
-    value: 0
+    value: '{ size.0.value }'
   tablet:
     value: 740px
   desktop:
     value: 992px
```
matthewferry, line 7: These are a little non-standard... with the new primitive scale based on our 4px scaling, we could use 720 for tablet and 960 for desktop (or 1120). But we can also keep these as is if we'd rather not mess with the responsive sizes.

matthewferry: or probably better idea to instead leave these as hard coded pixel values for now and we can look at breakpoints separately

Niznikr: No strong preference either way. wdyt @apucacao since you worked on breakpoints a bit for the product?

apucacao: i think leaving them as-is if it doesn't impede our progress with the new design system stuff here makes sense to me.

matthewferry: Cool let's leave as defined for now :)

packages/tokens/src/border.yaml
```diff
@@ -1,13 +1,13 @@
...
     medium:
-      value: 0.375rem
+      value: '{ size.6.value }'
   width:
     100:
-      value: 0rem
+      value: '{ size.0.value }'
     200:
-      value: 0.063rem
+      value: '{ size.0.value }'
```
Niznikr, line 11:
```suggestion
      value: '{ size.1.value }'
```
Or was it intentionally the same as `100`?

matthewferry: oh jeez, nice catch!

Niznikr (approved):

PR #1038, opened 2023-10-13
feat(tokens, icons): add new tokens and icons

vroske-ld:
## Summary

- Add new background, border, and color alias tokens for `success`
- Add a new border radius token
- Add gradient tokens
- A few tweaks to existing `flair` tokens
- Add `party-popper` and `stars` icons (needs to be optimized)


packages/tokens/src/color-primitives.yaml
```diff
@@ -189,3 +189,14 @@ color:
         value: hsl(45.1, 57.4%, 38.6%)
       900:
         value: hsl(45, 57.8%, 32.5%)
+  gradient:
+    yellowToCyan:
```
Niznikr, line 193: We've been using kebab case for our names so perhaps `yellow-to-cyan` or `yellow-cyan` would work better?

packages/tokens/src/color-primitives.yaml
```diff
@@ -189,3 +189,14 @@ color:
         value: hsl(45.1, 57.4%, 38.6%)
       900:
         value: hsl(45, 57.8%, 32.5%)
+  gradient:
```
Niznikr, line 192: If possible, can we also have these render in the [global tokens story](https://github.com/launchdarkly/launchpad-ui/blob/main/packages/tokens/stories/tokens.stories.tsx#L58-L65)?

packages/tokens/src/color-primitives.yaml
vroske-ld: ah thanks for the catch, this was me just copy-pasta'ing 🙂 i'll update this

packages/tokens/src/color-aliases.yaml
```diff
@@ -71,6 +71,19 @@ color:
         hover:
           value: '{ color.purple.100.value }'
           dark: '{ color.purple.800.value }'
+      success:
+        base:
+          value: '{ color.system.green.100.value }'
+          dark: '##00DA7B26' # color.system.green.500.value at opacity 15%
```
Niznikr, line 77:
```suggestion
          dark: '#00DA7B26' # color.system.green.500.value at opacity 15%
```

Niznikr (approved):

PR #1067, opened 2023-10-31
feat: add box package

Niznikr:
## Summary

First pass at a box component using [rainbow-sprinkles](https://github.com/wayfair/rainbow-sprinkles). Nice that is allows us to use our `vars` package to map to our tokens. The built CSS is only `1.48 kB` thanks to dynamic inlining CSS custom properties.

Screenshot 2023-10-31 at 11 35 51 AM

Screenshot 2023-10-31 at 11 37 00 AM


apucacao:
@portpaw Looks like you and @Niznikr were on a similar wavelength.

portpaw:
I'd prefer to resolve differences between our two proposals before merging either ([other proposal for reference](https://github.com/launchdarkly/gonfalon/pull/31779)). There's a lot here that I like (specifically the style generation from our tokens), but there are a few elements of my branch that could or maybe should be pulled over, too.

@Niznikr Can we set up some time early next week to discuss?

packages/box/src/styles/rainbow-sprinkles.css.ts
```diff
@@ -0,0 +1,117 @@
...
+  : never;
+
+type GlobalKeys = FlattenObjectKeys<typeof global>;
+type GradientKeys = FlattenObjectKeys<typeof gradient>;
+type BackgroundKeys = FlattenObjectKeys<typeof bg>;
+type BorderKeys = FlattenObjectKeys<typeof border>;
+type FillKeys = FlattenObjectKeys<typeof fill>;
+type TextKeys = FlattenObjectKeys<typeof text>;
+
+const colors = flatten<typeof global, Record<GlobalKeys, string>>(global);
```
portpaw, line 20: A question about how we tokenize these. Do we want to use hyphen `-` delimiters rather than periods `.`? That would map more cleanly to how they're expressed in CSS.

packages/box/src/styles/rainbow-sprinkles.css.ts
```diff
@@ -0,0 +1,117 @@
...
+    marginX: ['marginLeft', 'marginRight'],
+    marginY: ['marginTop', 'marginBottom'],
+    mx: ['marginLeft', 'marginRight'],
+    my: ['marginTop', 'marginBottom'],
+  },
+});
+
+const colorProperties = defineProperties({
+  conditions: {
+    lightMode: { selector: ':root &, [data-theme="default"] &' },
```
portpaw, line 99: A small thing: Maybe this should just be `default` and `dark`, to match the names of the themes as they exist in other parts of the code?

packages/box/src/styles/rainbow-sprinkles.css.ts
```diff
@@ -0,0 +1,117 @@
...
+    mt: ['marginTop'],
+    mb: ['marginBottom'],
+    marginX: ['marginLeft', 'marginRight'],
+    marginY: ['marginTop', 'marginBottom'],
+    mx: ['marginLeft', 'marginRight'],
+    my: ['marginTop', 'marginBottom'],
+  },
+});
+
+const colorProperties = defineProperties({
```
portpaw, line 97: Should this be `themedProperties` and maybe include `borderStyle` as well? Not sure if that would be considered a themed value or a responsive value 🤔 I made it themed in my implementation, but I wasn't 100% confident in that decision.

packages/box/src/styles/rainbow-sprinkles.css.ts
```diff
@@ -0,0 +1,117 @@
...
+  },
+  defaultCondition: 'lightMode',
+  dynamicProperties: {
+    color: { ...colors, ...texts },
+    background: gradients,
+    backgroundColor: { ...colors, ...backgrounds },
+    borderColor: { ...colors, ...borders },
+    fill: { ...colors, ...fills },
+  },
+  shorthands: {
```
portpaw, line 110: Same question here re: shorthand.

packages/box/src/Box.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+import { rainbowSprinkles } from './styles/rainbow-sprinkles.css';
+
+type BoxProps = Sprinkles & {
+  children?: ReactNode;
+  asChild?: boolean;
+  'data-test-id'?: string;
+};
+
+const Box = ({ asChild, children, 'data-test-id': testId = 'box', ...props }: BoxProps) => {
```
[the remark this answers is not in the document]
Niznikr: Perhaps only set it if `asChild` is not present? We typically set a default test ID for all LP components.

packages/box/src/styles/rainbow-sprinkles.css.ts
portpaw: Cool, that makes sense to me!

packages/box/src/styles/rainbow-sprinkles.css.ts
```diff
@@ -0,0 +1,117 @@
...
+    lineHeight: vars['line-height'],
+    zIndex: vars['z-index'],
+    width: vars.size,
+    height: vars.size,
+    // Will work with any CSS value
+    display: true,
+    textAlign: true,
+    flexDirection: true,
+    justifyContent: true,
+    alignItems: true,
```
portpaw, line 67: Another thing that's probably worth discussing: do we want flex properties to be part of a `Flex` component, rather than `Box` being overloaded?

packages/box/src/styles/rainbow-sprinkles.css.ts
```diff
@@ -0,0 +1,117 @@
...
+    marginTop: vars.spacing,
+    marginLeft: vars.spacing,
+    marginRight: vars.spacing,
+    marginBottom: vars.spacing,
+    borderRadius: vars.border.radius,
+    borderWidth: vars.border.width,
+    fontFamily: vars.font.family,
+    fontSize: vars.font.size,
+    fontWeight: vars.font.weight,
+    lineHeight: vars['line-height'],
```
portpaw, line 58: Same thing here: do we want to expose these properties on the `Box` component, or would we rather these concerns belong to a `Text` component?

packages/box/src/styles/rainbow-sprinkles.css.ts
Niznikr: I think a `Flex` and `Text` component will likely be something we want as a follow-up to rein things in and offer precise dialing. Since Box is meant polymorphic by definition we should "allow it all" for quick prototyping and versatility. I imagine Box as a base class for other components/packages (Flex/Text) to extend/inherit from.

portpaw (approved):

apucacao (approved):

PR #1124, opened 2024-01-04
refactor(components): use vanilla-extract for styles

Niznikr:
## Summary

While StyleX provides some nice optimizations via atomic CSS, a couple drawbacks became apparent when trying to create styles for buttons: debugging the long list of classes via the browser dev tools is a pain (common for all atomic CSS solutions) and there isn't a way to target ancestor selectors (in this case dark mode via data-theme attr) for styles. Vanilla Extract does not have either of these issues and makes it easy to define locally scoped CSS custom properties. I believe VE's versatility and wider feature set at the expense of a larger CSS bundle size is a better fit for our needs in the long run.


erangeles (approved):

packages/vars/stories/Story.css.ts
```diff
@@ -16,6 +16,11 @@ export const local = style({
   vars: {
     [accentVar]: vars.color.pink[500],
   },
+  selectors: {
+    ['[data-theme="dark"] &']: {
+      vars: { [accentVar]: vars.color.pink[200] },
+    },
+  },
```
Niznikr, line 23: Simple example of how to tweak a local CSS custom property (component-level token) for dark mode.

PR #1132, opened 2024-01-11
feat(components): add `Button`, `ButtonGroup`, `IconButton`, and `LinkButton`

Niznikr:
## Summary

Add `Button`, `ButtonGroup`, `IconButton`, and `LinkButton` RAC-based components.

### DIFF
--- a/packages/components/src/Button.tsx
+++ b/packages/components/src/Button.tsx
@@ -1,0 +1,24 @@
+import type { ButtonVariants } from './styles/Button.css';
+import type { ForwardedRef } from 'react';
+import type { ButtonProps as AriaButtonProps } from 'react-aria-components';
+
+import { clsx } from 'clsx';
+import { forwardRef } from 'react';
+import { Button as AriaButton } from 'react-aria-components';
+
+import { base, variants } from './styles/Button.css';
+
+type ButtonProps = AriaButtonProps & ButtonVariants;
+
+const Button = forwardRef(
+  ({ size = 'medium', className, ...props }: ButtonProps, ref: ForwardedRef<HTMLButtonElement>) => {
+    return (
+      <AriaButton {...props} ref={ref} className={clsx(base, variants({ size }), className)} />
+    );
+  }
+);
+
+Button.displayName = 'Button';
+
+export { Button };
+export type { ButtonProps };
--- a/packages/components/src/Button.tsx
+++ b/packages/components/src/Button.tsx
@@ -6,14 +6,17 @@ import { clsx } from 'clsx';
 import { forwardRef } from 'react';
 import { Button as AriaButton } from 'react-aria-components';
 
-import { base, variants } from './styles/Button.css';
+import { variants } from './styles/Button.css';
 
 type ButtonProps = AriaButtonProps & ButtonVariants;
 
 const Button = forwardRef(
-  ({ size = 'medium', className, ...props }: ButtonProps, ref: ForwardedRef<HTMLButtonElement>) => {
+  (
+    { size = 'medium', variant = 'default', className, ...props }: ButtonProps,
+    ref: ForwardedRef<HTMLButtonElement>
+  ) => {
     return (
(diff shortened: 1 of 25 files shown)


packages/components/src/styles/Button.css.ts
```diff
@@ -0,0 +1,195 @@
...
+        borderRight: 'none',
+      },
+      [`${buttonGroup.classNames.variants.spacing.compact} &:last-child`]: {
+        borderTopLeftRadius: 0,
+        borderBottomLeftRadius: 0,
+        borderLeft: 'none',
+      },
+      [`${buttonGroup.classNames.variants.spacing.compact} &:not(:first-child):not(:last-child)`]: {
+        borderRadius: 0,
+      },
```
Niznikr, line 46: Example of [circular selectors](https://vanilla-extract.style/documentation/styling/#circular-selectors)

packages/components/src/ButtonGroup.tsx
```diff
@@ -0,0 +1,31 @@
...
+import { ButtonContext } from 'react-aria-components';
+
+import { variants } from './styles/ButtonGroup.css';
+
+type ButtonGroupProps = ComponentPropsWithRef<'div'> &
+  ButtonGroupVariants & {
+    isDisabled?: boolean;
+  };
+
+const ButtonGroup = forwardRef(
```
apucacao, line 15: Curious: should we carry this group component forward?

The RAC-based components represent our pristine future, and IIRC ButtonGroup as it is today isn't one we're in love with.

packages/components/src/styles/Button.css.ts
```diff
@@ -0,0 +1,195 @@
+import type { RecipeVariants } from '@vanilla-extract/recipes';
+
+import { vars } from '@launchpad-ui/vars';
+import { recipe } from '@vanilla-extract/recipes';
+
+import { variants as buttonGroup } from './ButtonGroup.css';
+
+const variants = recipe({
+  base: {
```
apucacao, line 9: Just a personal thing, nothing to do: I do love my CSS, especially after seeing this. 😅

packages/components/src/ButtonGroup.tsx
Niznikr: I'd say yes if only for the styles for grouping them (which have been cleaned up) as it is a pattern we'd want consistency around. They even [have an example](https://react-spectrum.adobe.com/react-aria/Button.html#advanced-customization) for one. Can always drop it if we want to move a different direction later.

apucacao: Oh I missed you built the grouping on top of RAC!

Ignore me.

apucacao (approved):

packages/components/src/Button.tsx
```diff
@@ -0,0 +1,27 @@
...
+import type { ForwardedRef } from 'react';
+import type { ButtonProps as AriaButtonProps } from 'react-aria-components';
+
+import { clsx } from 'clsx';
+import { forwardRef } from 'react';
+import { Button as AriaButton } from 'react-aria-components';
+
+import { variants } from './styles/Button.css';
+
+type ButtonProps = AriaButtonProps & ButtonVariants;
```
apucacao, line 11: Oh, in launchpad-experimental I removed `disabled` to avoid confusing with `isDisabled` which is how RAC controls that state.

Curious what you think about that.

packages/components/src/styles/Button.css.ts
Niznikr: Is it at least easier to parse than the current button styles? This component probably has the most variants but I want to make sure VE doesn't intimidate folks too much.

packages/components/src/Button.tsx
Niznikr: Yeah same here. We should keep additional props to a minimum, and since folks can iteratively adopt we don't need to shy away from breaking changes/prop cleanup.

PR #1136, opened 2024-01-16
refactor(components): use CSS modules and Lightning CSS for styles

Niznikr:
## Summary

Use CSS modules and [Lightning CSS](https://lightningcss.dev/) for styles.


packages/components/src/Button.tsx
```diff
@@ -1,27 +1,49 @@
...
+      primaryFlair: styles.primaryFlair,
+      defaultFlair: styles.defaultFlair,
+      minimalFlair: styles.minimalFlair,
+    },
+  },
+  defaultVariants: {
+    size: 'medium',
+    variant: 'default',
+  },
+});
```
Niznikr, line 32: [cva](https://cva.style/docs) provides us a similar variant management DX that VE recipes did

erangeles (approved):
looks good to me, thanks @Niznikr !

PR #1148, opened 2024-01-25
feat(components): add `Switch`

Niznikr:
## Summary

Add `Switch` and add JSDoc to components so folks can easily access RAC docs.

### DIFF
--- a/packages/components/src/styles/Switch.module.css
+++ b/packages/components/src/styles/Switch.module.css
@@ -7,23 +7,19 @@
 .track {
   display: flex;
   align-items: center;
+  gap: var(--lp-spacing-200);
   cursor: pointer;
   padding: var(--lp-spacing-200);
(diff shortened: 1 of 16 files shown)


apucacao (approved):

packages/components/src/styles/Switch.module.css
```diff
@@ -0,0 +1,73 @@
...
+}
+
+.switch[data-selected] {
+  & .track {
+    background-color: var(--lp-color-system-green-700);
+    border: 1px solid var(--lp-color-system-green-800);
+  }
+
+  & .handle {
+    transform: translateX(calc(100cqw - 100%));
```
Niznikr, line 61: https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_container_queries#container_query_length_units

PR #1152, opened 2024-01-30
feat(components): add `Modal` and `ModalOverlay`

Niznikr:
## Summary

Add `Modal` with variants `default` and `drawer` since functionally they are the same.

https://github.com/launchdarkly/launchpad-ui/assets/2147624/346ea752-afa3-406c-b235-424ea007b399


apucacao (approved):

packages/components/src/styles/Modal.module.css
```diff
@@ -0,0 +1,201 @@
...
+  outline: none;
+  padding: var(--lp-spacing-700);
+  box-shadow: 0 0 6px 0 #939598;
+
+  & [slot='title'] {
+    font: var(--lp-heading-2-medium);
+    flex-grow: 1;
+  }
+
+  & [role='dialog']:has(> [slot='body']) {
```
apucacao, line 64: This bit is a bit risky IMO: what if consumer wants/needs to add something around the body slot? Is the `>` necessary?

Niznikr: Let me see if I can tweak that. Just need a check to make sure to only apply those styles if it has the slot present.

PR #1158, opened 2024-02-02
fix(components): align menu item contents

Niznikr:
## Summary

Add story for items with icons and menu item states.


pheggeseth (approved):
LGTM!

PR #1166, opened 2024-02-06
feat(components): add `RadioGroup`, `TextField`, `Group`, and `SearchField`

Niznikr:
## Summary

Add `RadioGroup`, `TextField`, `Group`, and `SearchField`.


packages/components/src/styles/Group.module.css
```diff
@@ -0,0 +1,19 @@
+.group {
+  composes: input from './Input.module.css';
```
Niznikr, line 2: [composes](https://lightningcss.dev/css-modules.html#class-composition) ftw. Going to refactor our focus states to utilize this.

PR #1167, opened 2024-02-07
feat: update bg and border tokens 

matthewferry:
- Creates consistency between button and form borders
- Creates consistency in hierarchy of default and dark mode
- Slightly softens our `black` token from true `#000000` to `#0B0B0B`

**Note:** while this change is not technically a breaking change because the API for the tokens has not changed at all, the color values have shifted slightly, so it may be necessary to change a few usages of bg-primary and bg-tertiary in gonfalon.


pheggeseth:
@matthewferry @apucacao @Niznikr I noticed that this is still active. Are we planning to merge this?

Niznikr:
@pheggeseth it needs rework but we haven't gotten to it

pheggeseth:
> @pheggeseth it needs rework but we haven't gotten to it

Cool sounds good. Just kept seeing this in my "needs review" feed and wanted to make sure that it didn't fall through the cracks or anything.

matthewferry:
Yeah apologies. It got out on hold. I'll get to this tomorrow, but in future, would it be better to close the PR and reopen when ready?

apucacao:
@matthewferry What's holding this one up now?

PR #1186, opened 2024-02-28
feat(components): use light-dark function

Niznikr:
## Summary

Use [light-dark() color function](https://lightningcss.dev/transpilation.html#light-dark()-color-function) to minimize redundancy and avoid custom property bloat.


apucacao (approved):
Is the idea that we use this instead of component-specific theming tokens? Or, how/when should we use this approach?

pheggeseth (approved):
LGTM. TIL about `color-scheme`.

PR #1192, opened 2024-03-05
fix(components): override `forwardRef` to support generics

Niznikr:
## Summary

RAC does something similar [via a type](https://github.com/adobe/react-spectrum/blob/main/packages/react-aria-components/src/utils.tsx#L19-L24) but I prefer [this approach](https://www.totaltypescript.com/forwardref-with-generic-components#the-solution).


pheggeseth:
Thanks!

PR #1209, opened 2024-03-21
feat(components): set `Group` context within `TextField`

Niznikr:
## Summary

Set `Group` context within `TextField` ([like SearchField does](https://github.com/adobe/react-spectrum/blob/main/packages/react-aria-components/src/SearchField.tsx#L100)) using [Provider](https://react-spectrum.adobe.com/react-aria/advanced.html#provider) to improve DX.


pheggeseth (approved):
Thanks!

PR #1211, opened 2024-03-21
fix: ensure RAC portals are included in axe scans

Niznikr:
## Summary

Ensure RAC portals are included in axe scans by using the `[data-rac]` selector that every RAC renders.

## Testing approaches

Run the Playwright test in [UI mode](https://playwright.dev/docs/test-ui-mode) and log the nodes from the `passes` field within the scan results.


matthewferry:
@erangeles is this the fix for the bug we've been seeing in IA?

PR #1212, opened 2024-03-22
fix(components): ensure `Tooltip` content can be hovered

Niznikr:
## Summary

Ensure `Tooltip` content can be hovered to meet [a11y requirements](https://www.w3.org/WAI/WCAG21/Understanding/content-on-hover-or-focus.html).

## Testing approaches

Open a tooltip and notice you can hover to the content without it closing.

### DIFF
--- a/packages/components/src/Tooltip.tsx
+++ b/packages/components/src/Tooltip.tsx
@@ -23,7 +23,7 @@ const _Tooltip = (props: TooltipProps, ref: ForwardedRef<HTMLDivElement>) => {
 	return (
 		<AriaTooltip
 			{...props}
-			offset={0}
+			offset={4}
 			crossOffset={0}
 			ref={ref}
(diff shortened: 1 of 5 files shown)


packages/components/src/Tooltip.tsx
```diff
@@ -41,7 +41,7 @@ const _Tooltip = (props: TooltipProps, ref: ForwardedRef<HTMLDivElement>) => {
 const Tooltip = forwardRef(_Tooltip);
 
 const TooltipTrigger = (props: TooltipTriggerProps) => {
-	return <AriaTooltipTrigger {...props} delay={500} closeDelay={0} />;
+	return <AriaTooltipTrigger {...props} delay={500} closeDelay={250} />;
```
Niznikr, line 44: RAC default is 500 and we had set it to 0 resulting in the issue.

pheggeseth (approved):

PR #1218, opened 2024-03-27
feat(components): add `LinkIconButton`

Niznikr:
## Summary

Add `LinkIconButton` which are links that look like icon buttons.


packages/components/src/LinkIconButton.tsx
```diff
@@ -0,0 +1,49 @@
...
+import { cx } from 'class-variance-authority';
+import { forwardRef } from 'react';
+import { composeRenderProps } from 'react-aria-components';
+
+import { button } from './Button';
+import { iconButton } from './IconButton';
+import { Link } from './Link';
+
+interface LinkIconButtonProps
+	extends Omit<LinkProps, 'variant' | 'aria-label'>,
```
apucacao, line 17: Why are we removing variant and aria-label then re-adding them?

apucacao (approved):

packages/components/src/LinkIconButton.tsx
Niznikr: That helper seems to use intersections which I rather avoid given they've caused issues here. I'll stick with interfaces and native TS utilities.

Niznikr: image
Agreed. I just don't think they make sense in LaunchPad given they recursively merge props and mask potential conflicts.

apucacao: What I'm getting at is this: I understand there are differences between interface and type, and that generally we want to use interface for our (prop?) types.

But TypeScript itself ships with `type`-based utils, and it's still a useful feature used correctly.

We shouldn't necessarily discard an entire part of the language because of the issue we ran into, unless we have good reason to believe that those features shouldn't even be in the language itself, and therefore should be avoided at all cost.

Niznikr: No disagreements there. To minimize risk and bugs for this project I'd rather limit usage here. Both can be true.

PR #1246, opened 2024-04-16
fix(components): extract `Header` styles

Niznikr:
## Summary

Extract `Header` styles out from `Menu`.

### DIFF
--- a/packages/components/src/Header.tsx
+++ b/packages/components/src/Header.tsx
@@ -1,3 +1,20 @@
-import { Header } from 'react-aria-components';
+import type { ForwardedRef, HTMLAttributes } from 'react';
+
+import { cva } from 'class-variance-authority';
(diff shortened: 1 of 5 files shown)


packages/components/src/Header.tsx
```diff
@@ -1,3 +1,20 @@
...
+import { cva } from 'class-variance-authority';
+import { forwardRef } from 'react';
+import { Header as AriaHeader } from 'react-aria-components';
+
+import styles from './styles/Header.module.css';
+
+const header = cva(styles.header);
+
+const _Header = (
+	{ className, ...props }: HTMLAttributes<HTMLElement>,
```
pheggeseth, line 12: TIL that there is no interface for a `header` element, nor for any of the other sectioning elements, even though there is one for `div`!

pheggeseth (approved):

PR #1252, opened 2024-04-18
fix(components): remove border from `Popover`

Niznikr:
## Summary

Remove border from `Popover`.


pheggeseth:
I'm just curious as to why? I personally find that having a subtle border for a popover helps more crisply define where the overlay is, especially in dark mode.

Niznikr:
Preference from design. I believe they are going to revisit the box shadows and improve them.

PR #1254, opened 2024-04-19
fix(components): fix `Checkbox` label alignment when wrapped

Niznikr:
## Summary

Fix `Checkbox` label alignment when wrapped by putting it in a container with height set to `1lh` to match the label's.

## Testing approaches

Modify children of Checkbox in Storybook until the label wraps and notice the alignment is correct.


hisuida (approved):
Thank you! :D

PR #1266, opened 2024-04-25
feat(components): add `ToggleButton`

Niznikr:
## Summary

Add `ToggleButton`.


apucacao (approved):
Do you have _any_ idea how often I've wanted one of these recently?? 😆

PR #1275, opened 2024-05-01
refactor(components): inset focus outlines

Niznikr:
## Summary

Inset focus outlines to avoid issues with overflow settings on containers cutting them off.


apucacao (approved):
Amazing, thank you!

PR #1307, opened 2024-05-17
fix(icons)!: remove package-send-bolt and update package-send

matthewferry:
## Summary

Updates the package send icon to sit better optically and removes the package-send-bolt icon which is currently unused anyway


.changeset/rare-ears-rhyme.md
```diff
@@ -0,0 +1,6 @@
+---
+"@launchpad-ui/icons": minor
+"@launchpad-ui/core": minor
```
Niznikr, line 3:
```suggestion
```
Changesets will automatically update `core` and nothing breaks from a core's usage perspective

matthewferry: ahhhh, gotcha. Last time I read the CONTRIBUTING docs, it had the line about _always_ including core. but I see that's since been removed! 

https://github.com/launchdarkly/launchpad-ui/commit/bdd6f5c24d9d4f1237e3269cb917b5b935bdd89f

PR #1321, opened 2024-05-31
feat(components): add `Calendar`, `DateField`, `DatePicker`, `DateRangePicker`, `RangeCalendar`, and `TimeField`

Niznikr:
## Summary

Add `Calendar`, `DateField`, `DatePicker`, `DateRangePicker`, `RangeCalendar`, and `TimeField`. To follow up with [preset dates functionality](https://react-spectrum.adobe.com/react-aria/Calendar.html#contexts).

## Testing approaches

Check stories in `Date and Time` section.

### DIFF
--- a/packages/components/__tests__/DateField.spec.tsx
+++ b/packages/components/__tests__/DateField.spec.tsx
@@ -1,7 +1,7 @@
 import { describe, expect, it } from 'vitest';
 
 import { render, screen } from '../../../test/utils';
-import { DateField, DateInput, DateSegment, Label } from '../src';
+import { DateField, DateInput, DateSegment, Label, TimeField } from '../src';
 
 describe('DateField', () => {
 	it('renders', () => {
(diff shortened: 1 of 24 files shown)


apucacao:
Small UX thing:
- The react-aria-components instances are styled in a way that hides the (text) cursor, but which feels and looks great: https://react-spectrum.adobe.com/react-aria/DatePicker.html
- In storybook, the current implement shows a cursor for the relevant component (like the year), but it doesn't "feel" like a normal cursor: it sticks to the beginning of the year field.

Did you guys consider styling the inputs in a way similar to how RAC does it? I think it would give the components a more solid feel.

packages/components/__tests__/DateField.spec.tsx
```diff
@@ -0,0 +1,28 @@
+import { describe, expect, it } from 'vitest';
+
+import { render, screen } from '../../../test/utils';
+import { DateField, DateInput, DateSegment, Label, TimeField } from '../src';
```
apucacao, line 4: I think I missed the `DateSegment` PR, my bad.

We should add `font-variant-numeric: tabular-nums;` to the segments so that their width doesn't shift ever-so-slightly as the numbers change.

apucacao (approved):

packages/components/__tests__/DateField.spec.tsx
Niznikr: Design wasn't a fan of the kerning with tabular-nums. Perhaps something worth following up on to see if we can achieve it with reduced letter spacing.

PR #1325, opened 2024-06-05
refactor(tokens)!: rework typography tokens

Niznikr:
## Summary

- Remove font-size-0
- font-size-100: 12px -> 11px
- font-size-200: 14px -> 13px
- font-size-300: 16px -> 15px
- Update line-height tokens
- Remove small-2-*


packages/tokens/src/font.json
```diff
@@ -13,32 +13,29 @@
...
-			"$value": "{size.24}"
+			"$value": "{size.32}"
 		},
 		"700": {
-			"$value": "{size.32}"
+			"$value": "{size.44}"
 		},
 		"800": {
-			"$value": "{size.44}"
+			"$value": "3.75rem"
```
vroske-ld, line 38: is there a reason this doesn't have a corresponding `size` token?

packages/counter/src/styles/Counter.module.css
```diff
@@ -1,8 +1,9 @@
+/* stylelint-disable number-max-precision */
 .Counter {
   display: inline-block;
   font-size: 0.6875rem;
   font-weight: var(--lp-font-weight-bold);
-  line-height: var(--lp-line-height-200);
+  line-height: 1.285714286;
```
vroske-ld, line 6: why do we have this very specific number here?

Niznikr: The old value for `lp-line-height-200`

packages/tokens/src/font.json
Niznikr: Tentative value so we wanted to wait before adding it as a size.

vroske-ld (approved):

PR #1331, opened 2024-06-10
feat(components): add `subtitle` slot to `Dialog`

Niznikr:
## Summary

Add `subtitle` slot to `Dialog` that describes modals.


packages/components/stories/Modal.stories.tsx
```diff
@@ -57,6 +66,7 @@ const renderModal = (args: Story['args'], destructive = false) => (
 									variant="minimal"
 									onPress={close}
 								/>
+								<Text slot="subtitle">Subtitle</Text>
```
apucacao, line 69: This is so amazing.

apucacao (approved):

PR #1383, opened 2024-08-14
feat: add dist files to tokens package exports

kwatkins-ld:
## Summary

For some bundlers (e.g. Webpack), the file-system is the source-of-truth and so imports like:
```css
@import '@launchpad-ui/tokens/dist/media-queries.css';
```
are supported because the `media-queries.css` file is present on the file system in that location. Other bundlers (e.g. Vite) honor the `exports` map in a `package.json` file, so the same import breaks:
```css
/* Vite will error, because this file isn't listed in the exports map in package.json */
@import '@launchpad-ui/tokens/dist/media-queries.css';
```

In the case of the `@launchpad-ui/tokens` package was incomplete (we are including the "virtual" exports, but not the actual files on disk in the `dist` folder).

### DIFF
--- a/packages/tokens/package.json
+++ b/packages/tokens/package.json
@@ -29,6 +29,9 @@
 		"./themes.css": "./dist/themes.css",
 		"./index.css": "./dist/index.css",
 		"./media-queries.css": "./dist/media-queries.css",
+		"./dist/themes.css": "./dist/themes.css",
+		"./dist/index.css": "./dist/index.css",
+		"./dist/media-queries.css": "./dist/media-queries.css",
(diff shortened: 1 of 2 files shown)


packages/tokens/package.json
```diff
@@ -29,6 +29,9 @@
 		"./themes.css": "./dist/themes.css",
 		"./index.css": "./dist/index.css",
 		"./media-queries.css": "./dist/media-queries.css",
```
kwatkins-ld, line 31: I considered just updating the existing "virtual" exports above, but that would be a breaking change (open to feedback though!)

Niznikr (approved):

PR #1405, opened 2024-09-18
feat(tokens): include fonts in package

Niznikr:
## Summary

Include fonts in `tokens` package.


apucacao (approved):
Just curious: is this happening because of the issues folks ran into with LaunchDevly's UI?

Niznikr:
Mainly for convenience, but it can also reduce the chance folks run into inconsistencies.

PR #1406, opened 2024-09-19
fix: ensure proper overlay background layering

Niznikr:
## Summary

Ensure proper overlay background layering.

image-20240918-185348

### DIFF
--- a/packages/tokens/src/color-aliases.json
+++ b/packages/tokens/src/color-aliases.json
@@ -183,10 +183,10 @@
 			"overlay": {
 				"primary": {
 					"$value": "{color.white.950}",
-					"dark": "{color.gray.900}"
+					"dark": "{color.gray.950}"
 				},
 				"secondary": {
-					"$value": "{color.gray.950}",
+					"$value": "{color.white.900}",
 					"dark": "{color.gray.900}"
 				}
 			},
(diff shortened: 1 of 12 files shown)


apucacao (approved):
That's an Apple  quality graphic.

kwatkins-ld (approved):

packages/tokens/src/color-aliases.json
```diff
@@ -183,10 +183,10 @@
 			"overlay": {
 				"primary": {
 					"$value": "{color.white.950}",
-					"dark": "{color.gray.900}"
+					"dark": "{color.gray.950}"
 				},
 				"secondary": {
-					"$value": "{color.gray.950}",
+					"$value": "{color.white.900}",
```
matthewferry, line 189: hm, white 900 is semi transparent. I think maybe we mean to use `gray.0`? tho I'm not actually sure where we use secondary overlay tokens?

Niznikr: Updated `Popover` to use secondary overlay since it should use its dark value.

Niznikr: So had the value mirror how primary has white 950 and gray 950.

Niznikr:
Closing in favor of the color refresh which will update the bg ui tokens to replace the overlay ones.

PR #1411, opened 2024-09-24
refactor(icons)!: align API with RAC

Niznikr:
## Summary

Align API with RAC components:
  - Remove `IconContext`
  - Remove outer `span`
  - Remove props `title` and `description` in favor of composition
  - Prop `subtle` -> `variant`
  - Prop `children` for custom icons
  - Consolidate sizes


.storybook/preview.tsx
```diff
@@ -16,6 +18,32 @@ import '../packages/tokens/dist/index.css';
 import '../packages/tokens/dist/media-queries.css';
 import '../packages/tokens/dist/themes.css';
 
+fetch(sprite)
```
kwatkins-ld, line 21: Not a problem for today, but I wonder if we should eventually wrap this logic up in a helper and export it (we have identical setup code in consumers).

kwatkins-ld (approved):

.storybook/preview.tsx
Niznikr: Yeah that's a good idea 😄

PR #1416, opened 2024-09-27
feat(icons): add `data-icon` attribute

Niznikr:
## Summary

Add `data-icon` attribute so consumers can easily target icons by name.


apucacao (approved):
So considerate! Seriously though, this is awesome.

PR #1432, opened 2024-10-04
feat(box): support `className`

Niznikr:
## Summary

Support `className`.


packages/box/src/Box.tsx
```diff
@@ -1,27 +1,21 @@
...
-	'data-test-id': testId = asChild ? undefined : 'box',
-	...props
-}: BoxProps) => {
+const Box = ({ asChild, children, className: classes, ...props }: BoxProps) => {
 	const Component = asChild ? Slot : 'div';
 	const { className, style, otherProps } = rainbowSprinkles(props);
 
 	return (
-		<Component className={className} style={style} data-test-id={testId} {...otherProps}>
+		<Component className={cx(classes, className)} style={style} {...otherProps}>
```
zmdavis, line 18: Personally I find this a little confusing and would prefer to rename the sprinkles variable to make it more obvious what's happening, but that's minor and you can take or leave that suggestion.

zmdavis (approved):

PR #1440, opened 2024-10-10
fix(button): update font styles to match RAC

Niznikr:
## Summary

Update font styles to match RAC.


apucacao (approved):

zmdavis (commented):
Thank you for this!

PR #1446, opened 2024-10-16
fix(components): use separate stacking context for `Switch`

Niznikr:
## Summary

https://developer.mozilla.org/en-US/docs/Web/CSS/isolation


zmdavis (approved):
Fun, I wasn't aware of this. Thanks for the quick fix!

PR #1452, opened 2024-10-23
feat(components): add `orientation` prop to `ButtonGroup`

Niznikr:
## Summary

Add `orientation` prop to `ButtonGroup`.

### DIFF
--- a/packages/components/src/styles/ButtonGroup.module.css
+++ b/packages/components/src/styles/ButtonGroup.module.css
@@ -7,27 +7,18 @@
 }
 
 .compact {
+	composes: buttonGroup from './base.module.css';
 	gap: initial;
-
-	& :first-child {
-		border-top-right-radius: 0;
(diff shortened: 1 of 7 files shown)


packages/components/src/styles/ButtonGroup.module.css
```diff
@@ -7,27 +7,18 @@
 }
 
 .compact {
+	composes: buttonGroup from './base.module.css';
```
pheggeseth, line 10: whoa is this a general feature of CSS Modules?!

pheggeseth (approved):

packages/components/src/styles/ButtonGroup.module.css
Niznikr: Yup! https://github.com/css-modules/css-modules/blob/master/docs/composition.md

apucacao (approved):

PR #1504, opened 2024-12-05
feat(components): add `TagButton`

Niznikr:
## Summary

Add `TagButton` to support tooltips/popovers until RAC natively supports tooltips on collections.


erangeles (approved):
🚀

PR #1508, opened 2024-12-11
feat(components): add `warning` variant for `Alert`

Niznikr:
## Summary

Add `warning` variant for `Alert`.

### DIFF
--- a/packages/components/src/styles/Alert.module.css
+++ b/packages/components/src/styles/Alert.module.css
@@ -19,6 +19,10 @@
 	&.success .icon {
 		fill: var(--lp-color-fill-feedback-success);
 	}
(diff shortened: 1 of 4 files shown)


packages/components/src/styles/Alert.module.css
```diff
@@ -47,6 +51,11 @@
 		--color-2: var(--lp-color-green-400);
 	}
 
+	&.warning {
+		--color-1: var(--lp-color-brand-orange-base);
+		--color-2: #ffb052;
```
Niznikr, line 56: To be replaced with new orange primitives later.

ld-kyee (approved):

PR #1519, opened 2025-01-10
feat(components): add `Avatar`

Niznikr:
## Summary

Add `Avatar`.

### DIFF
--- a/packages/components/src/Avatar.tsx
+++ b/packages/components/src/Avatar.tsx
@@ -1,8 +1,6 @@
-import type { IconProps } from '@launchpad-ui/icons';
 import type { VariantProps } from 'class-variance-authority';
 import type { ImgHTMLAttributes, RefObject, SVGAttributes } from 'react';
 
-import { Icon } from '@launchpad-ui/icons';
 import { cva, cx } from 'class-variance-authority';
 
 import styles from './styles/Avatar.module.css';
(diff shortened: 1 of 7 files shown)


packages/components/src/Avatar.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+	if (status === 'error') {
+		return children ? (
+			<InitialsAvatar size={size}>{children}</InitialsAvatar>
+		) : (
+			<IconAvatar size={size} />
+		);
+	}
+
+	// biome-ignore lint/a11y/useAltText: <explanation>
```
vezaynk, line 59: Will biome detect a violation if `Avatar` was not provided an `alt` tag? I assume its possible to configure it to do that (its possible with eslint).

packages/components/src/Avatar.tsx
```diff
@@ -0,0 +1,105 @@
...
+}
+
+interface IconAvatarProps extends Omit<IconProps, 'size'>, AvatarVariants {
+	ref?: RefObject<SVGSVGElement | null>;
+}
+
+interface InitialsAvatarProps extends SVGAttributes<SVGElement>, AvatarVariants {}
+
+const Avatar = ({ className, children, size = 'medium', ref, src, ...props }: AvatarProps) => {
+	const status = useImageLoadingStatus(src);
```
vezaynk, line 49: I'm wondering what your thoughts are on enabling this for Suspense. I imagine you would usually want to display a fallback (skeleton?) for when the Avatar isn't instantly available.

Niznikr: Perhaps I'll tweak the condition below to check `status !== 'loaded'` to render the fallback icon or initials version.

packages/components/src/Avatar.tsx
Niznikr: Yeah ignored since React types has it as optional `alt?: string | undefined`

ld-kyee (approved):
🎉

PR #1520, opened 2025-01-10
chore: update SB navigation and remove Box from BadgeIcon

matthewferry:
## Summary

- Updates the nav in storybook
- Removes Box from BadgeIcon
- Moves generic docs from /core package to docs dir
- Separates recipes into individual docs

new nav in sb

### DIFF
--- a/packages/components/stories/recipes/composition.stories.tsx
+++ b/packages/components/stories/recipes/composition.stories.tsx
@@ -40,6 +40,9 @@ const Container = (props: ComponentPropsWithoutRef<typeof Fragment>) => <>{props
 
 const meta: Meta<typeof Container> = {
 	component: Container,
+	parameters: {
+		chromatic: { pauseAnimationAtEnd: true },
+	},
 	argTypes: {
 		children: {
 			control: false,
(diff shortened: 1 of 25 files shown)


packages/components/stories/recipes/composition.stories.tsx
```diff
@@ -34,28 +34,18 @@ import {
...
-	parameters: {
-		chromatic: { pauseAnimationAtEnd: true },
-	},
-	decorators: [
-		(Story) => (
-			<div style={{ height: 'var(--lp-size-224)' }}>
-				<Story />
-			</div>
-		),
-	],
```
Niznikr, line 53: Need these to accurately snapshot overlays in Chromatic

matthewferry: How do I do that?

matthewferry: oh I see, I need to add the param back

matthewferry: Gotcha. Added back. Had ditched it because in storybook, absolutely positioned elements are seen and it looked better to just be the height of the initial in-the-document-flow content. Wish we could adjust the viewport size for chromatic snapshots, but not the story itself. bummer.

Niznikr (approved):

PR #1569, opened 2025-03-06
fix(components): pass label props to fallback `InitialsAvatar`

Niznikr:
## Summary

Pass label props to fallback `InitialsAvatar` using RA utils.

### DIFF
--- a/packages/components/src/Avatar.tsx
+++ b/packages/components/src/Avatar.tsx
@@ -1,6 +1,7 @@
 import type { VariantProps } from 'class-variance-authority';
 import type { ImgHTMLAttributes, Ref, SVGAttributes } from 'react';
 
+import { mergeProps, useLabels } from '@react-aria/utils';
 import { cva, cx } from 'class-variance-authority';
 
 import styles from './styles/Avatar.module.css';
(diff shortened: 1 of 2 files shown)


packages/components/src/Avatar.tsx
```diff
@@ -41,13 +42,25 @@ interface InitialsAvatarProps extends SVGAttributes<SVGElement>, AvatarVariants
...
 	if (status !== 'loaded') {
-		return <InitialsAvatar size={size}>{children}</InitialsAvatar>;
+		return (
+			<InitialsAvatar size={size} {...labelProps}>
+				{children}
+			</InitialsAvatar>
+		);
 	}
 
 	// biome-ignore lint/a11y/useAltText: <explanation>
```
kcacosta, line 55: @Niznikr, question about this comment above, this is the reason why it is ally test was failing (saying it was missing an alt text). why are we ignoring this lint?

kcacosta (approved):

packages/components/src/Avatar.tsx
Niznikr: It's looking for explicit setting of the alt text. Since this is an authored component and the label is passed implicitly from the props I just silenced it.

kcacosta: ah I understand, thanks for clarifying!

erangeles (approved):

PR #1577, opened 2025-03-10
feat: rel-6363 add arrow-upper-right-icon to launchpad

ari-launchdarkly:
## Summary

- Adding the arrow-upper-right-icon to launchpad icons based on the following figma designs:

https://www.figma.com/design/TOZXOu4NUhBZ6KxVJ0UW4U/branch/UUQXmbfQZLvMZXaTiVYEVr/FY26---Trial%2C-Onboarding?node-id=3312-35299&t=ENQzB9ZPQRul1uA4-0

Screenshot 2025-03-10 at 6 07 34 AM


.changeset/short-walls-read.md
```diff
@@ -0,0 +1,5 @@
+---
+"@launchpad-ui/icons": minor
```
Niznikr, line 2:
```suggestion
"@launchpad-ui/icons": patch
```
We can set to patch since [we follow 0.x versioning](https://github.com/launchdarkly/launchpad-ui/blob/main/CONTRIBUTING.md#versioning)

Niznikr (approved):

jahoni (commented):
lgtm

packages/icons/src/types.ts
```diff
@@ -27,6 +27,7 @@ const icons = [
 	'arrow-undo',
 	'arrow-up-right-circle',
 	'arrow-up-right-circle-outline',
+	'arrow-upper-right',
```
matthewferry, line 30: Nit, but can we rename this to `arrow-up-right`

matthewferry (changes requested):

packages/icons/src/types.ts
matthewferry: this follows the existing pattern e.g. arrow-up-right-circle which is this same icon in a circle

ari-launchdarkly: Ya totally. @matthewferry for future reference - is what's in figma the source of truth in terms of naming, or can they coexist as different names?

Niznikr (approved):

matthewferry (approved):

PR #1581, opened 2025-03-11
refactor(components)!: use RAC `Toast`

Niznikr:
## Summary

Use RAC `Toast` and export queues directly instead of an abstraction.

https://626696a2018c1f004a1cde86-mabprwbzaz.chromatic.com/?path=/docs/components-status-toast--docs


packages/components/src/Toast.tsx
```diff
@@ -57,197 +63,149 @@ interface RegionVariants extends VariantProps<typeof region> {}
...
+		RegionVariants {}
+
+const animate = (fn: () => void) => {
+	if ('startViewTransition' in document) {
+		document
+			.startViewTransition(() => {
+				flushSync(fn);
+			})
+			.ready.catch(() => {});
+	} else {
```
nhironaka, line 89: the spacing looks wonky here but I wonder if it's just github?

packages/components/src/Toast.tsx
```diff
@@ -57,197 +63,149 @@ interface RegionVariants extends VariantProps<typeof region> {}
...
+	extends Partial<AriaToastRegionProps<LPToastContent>>,
+		RegionVariants {}
+
+const animate = (fn: () => void) => {
+	if ('startViewTransition' in document) {
+		document
+			.startViewTransition(() => {
+				flushSync(fn);
+			})
+			.ready.catch(() => {});
```
nhironaka, line 88: do we need this line?

nhironaka (approved):

packages/components/src/Toast.tsx
Niznikr: Figured I'd match how the RAC folks implemented it https://github.com/adobe/react-spectrum/blob/main/packages/%40react-spectrum/toast/src/ToastContainer.tsx#L46

PR #1595, opened 2025-03-27
feat(components): export contexts

Niznikr:
## Summary

Export contexts to allow consumers to set props without conflicting with RAC.


packages/components/src/Button.tsx
```diff
@@ -36,35 +36,33 @@ const button = cva(styles.base, {
...
 });
 
-const ButtonContext = createContext<ButtonProps | null>(null);
-
 interface ButtonVariants extends VariantProps<typeof button> {}
 interface ButtonProps extends AriaButtonProps, ButtonVariants {
 	ref?: Ref<HTMLButtonElement>;
 }
 
+const ButtonContext = createContext<ContextValue<ButtonProps, HTMLButtonElement>>(null);
```
apucacao, line 44: Not sure if that rings a bell, but when testing this yesterday, `isPressed` (from their example) was causing a TypeScript error as if this version didn't support it? 🤔

apucacao (approved):
thanks for pushing this forward! 🙏

PR #1615, opened 2025-04-09
feat(components): add `bar` variant for `Meter` and `ProgressBar`

Niznikr:
## Summary

Add `bar` variant for `Meter` and `ProgressBar`.

### DIFF
--- a/packages/components/src/styles/ProgressBar.module.css
+++ b/packages/components/src/styles/ProgressBar.module.css
@@ -11,12 +11,12 @@
 .progress {
 }
 
-.circular {
+.spinner {
 	--fill-color: var(--lp-color-gray-500);
 	display: inline-flex;
 }
 
-.line {
+.bar {
 	--fill-color: var(--lp-color-brand-cyan-base);
 	display: flex;
 	flex-direction: column;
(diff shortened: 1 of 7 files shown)


apucacao (approved):

packages/components/src/styles/ProgressBar.module.css
```diff
@@ -9,16 +9,52 @@
...
 .innerCircle {
-	stroke: var(--fill);
+	stroke: var(--fill-color);
+}
+
+.container {
+	display: flex;
+}
+
+.label {
```
matthewferry, line 38: Do we need this? Looks like we encourage passing the `<Label/>` component in and that should already be styled to be consistent with other labels.

packages/components/src/styles/ProgressBar.module.css
```diff
@@ -9,16 +9,52 @@
...
+	flex: 1;
+	font: var(--lp-text-label-1-semibold);
```
matthewferry, line 40: If we do need this, then I'd suggest we align this with our other labels:
```suggestion
	font: var(--lp-text-label-1-medium);
```

matthewferry (approved):
Minor nit

packages/components/src/styles/ProgressBar.module.css
Niznikr: I guessed semibold from the design shared but we can just use the standard label.

PR #1619, opened 2025-04-10
fix(components): make sure button group only tweaks radii of immediate descendants

apucacao:

### DIFF
--- a/packages/components/src/styles/base.module.css
+++ b/packages/components/src/styles/base.module.css
@@ -10,41 +10,41 @@
 
 .buttonGroup {
 	&[data-orientation='horizontal'] {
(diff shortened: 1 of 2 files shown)


nhironaka (approved):

packages/components/src/styles/base.module.css
```diff
@@ -10,41 +10,41 @@
...
 
-		& :not(:first-child):not(:last-child) {
+		& > :not(:first-child):not(:last-child) {
 			border-radius: 0;
 		}
 	}
 
 	&[data-orientation='vertical'] {
-		& :first-child {
+		& > :first-child:not(:only-child) {
```
apucacao, line 33: Looks like this didn't quite work, according to Chromatic.

PR #1635, opened 2025-04-16
fix(components): remove flex styles from `Tooltip`

Niznikr:
## Summary

Remove flex styles from `Tooltip`. Originally added to align content with `kbd`. Consumers should instead be responsible for content layout.

### DIFF
--- a/packages/components/src/styles/Tooltip.module.css
+++ b/packages/components/src/styles/Tooltip.module.css
@@ -27,6 +27,7 @@
 	transform: translate3d(0, 0, 0);
 	will-change: transform, opacity;
 	max-width: var(--lp-size-400);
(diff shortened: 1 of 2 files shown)


packages/components/src/styles/Tooltip.module.css
```diff
@@ -17,10 +17,6 @@
 	isolation: isolate;
 	font: var(--lp-text-body-2-regular);
 	padding: var(--lp-spacing-200) var(--lp-spacing-300);
-	display: flex;
```
apucacao, line 20: Would you mind also adding `overflow: auto`? Generally, I think tooltips should be allowed to scroll?

apucacao (approved):

PR #1666, opened 2025-05-12
fix(components): fix button styles when popover is open

hsadhvani:
## Summary

Add button style for popover open state using aria-expanded and data-pressed

Popover on `Button`
Screenshot 2025-05-12 at 9 10 23 AM

Popover on `IconButton`
Screenshot 2025-05-12 at 9 10 34 AM

## Testing approaches

Manual 

---
Related Jira issue: [EXPT-2213: Popovers UI ](https://launchdarkly.atlassian.net/browse/EXPT-2213)

### DIFF
--- a/packages/components/src/styles/Button.module.css
+++ b/packages/components/src/styles/Button.module.css
@@ -1,5 +1,5 @@
 .base {
-	composes: interactive from "./base.module.css";
+	composes: interactive from './base.module.css';
 }
 
 .button {
@@ -161,9 +161,9 @@
 }
 
 .picker {
-	composes: base from "./Input.module.css";
-	composes: _default from "./Input.module.css";
-	composes: group from "./Group.module.css";
+	composes: base from './Input.module.css';
+	composes: _default from './Input.module.css';
+	composes: group from './Group.module.css';
 
 	& [data-icon]:last-child {
 		fill: var(--lp-color-fill-ui-secondary);
--- a/packages/components/src/styles/Button.module.css
+++ b/packages/components/src/styles/Button.module.css
@@ -139,6 +139,10 @@
 	z-index: 1;
 }
 
+.button[data-pressed][aria-expanded] {
+	border-color: var(--lp-color-border-interactive-selected);
+}
+
 .button[data-pending] {
 	cursor: wait;
 
--- a/packages/components/src/styles/Button.module.css
+++ b/packages/components/src/styles/Button.module.css
@@ -1,5 +1,5 @@
 .base {
-	composes: interactive from './base.module.css';
+	composes: interactive from "./base.module.css";
 }
 
 .button {
@@ -141,6 +141,7 @@
 
 .button[data-pressed][aria-expanded] {
 	border-color: var(--lp-color-border-interactive-selected);
+	background-color: var(--lp-color-bg-interactive-selected);
 }
 
 .button[data-pending] {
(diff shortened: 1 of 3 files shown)


packages/components/src/styles/Button.module.css
```diff
@@ -139,6 +139,11 @@
 	z-index: 1;
 }
 
+.button[data-pressed][aria-expanded] {
+	border-color: var(--lp-color-border-interactive-selected);
+	background-color: var(--lp-color-bg-interactive-selected);
```
apucacao, line 144: I'm 99% sure back when we had these styles, we only kept the background color. The border color is typically used for focus states, and in this case feel a bit too loud (especially considering that popovers are interactive and may contain their own focused children).

packages/components/src/styles/Button.module.css
```diff
@@ -139,6 +139,11 @@
 	z-index: 1;
 }
 
+.button[data-pressed][aria-expanded] {
```
apucacao, line 142: Weird, I swear we also used to have active/pressed styles.

Does `aria-expanded` _always_ come with the `data-pressed`?

packages/components/src/styles/Button.module.css
hsadhvani: @cchung-at-LD suggested we use the toggle button component instead of a button for popovers, as it has an on/off state to open/close the popover. So I copied over the styles used for the toggle button.

packages/components/src/styles/Button.module.css
hsadhvani: `aria-expanded` is for tooltips, and `aria-expanded` and `data-pressed` would be for popovers, as you need to press to open the popover.

packages/components/src/styles/Button.module.css
apucacao: I don't know that that's the right use case for toggle though.

@matthewferry do you and the LP folks have thoughts here?

hsadhvani: Well, that's why having some styles here helps. Although the border color might be a little too much here

apucacao:
FYI I spoke with @matthewferry and it sounds like he and @antonio-darkly have already stated that we should not include those active/pressed styles for now.

So unless we need a bigger convo, we should probably close this.

hsadhvani:
> FYI I spoke with @matthewferry and it sounds like he and @antonio-darkly have already stated that we should not include those active/pressed styles for now.
> 
> So unless we need a bigger convo, we should probably close this.

cc @cchung-at-LD

hsadhvani:
> 

But does that mean we don't want any active states for buttons with popovers?

matthewferry:
> But does that mean we don't want any active states for buttons with popovers?
yes. the popover is the active state

hsadhvani:
> > But does that mean we don't want any active states for buttons with popovers?
> 
> yes. the popover is the active state

I think you misunderstood. Chris wanted to use ToggleButtons for popovers, as buttons don't have an active state. 

But the intent of a ToggleButton is not to popovers but to toggle something on/off. So this PR intends to add an active state to a button that opens the popover as that's generally the use case for opening popovers.

matthewferry:
I'm not sure this is a pattern we should be adopting and this doesn't account for active states of buttons not using popovers or all button styles including minimal and primary buttons.

matthewferry:
wouldn't this affect tooltips as well?

packages/components/src/styles/Button.module.css
antonio-darkly: I agree with @apucacao here. Toggle Button isn't intended for open/close states. When you open a popover, a true/false state is overkill.

antonio-darkly:
Jumping into the conversations. @matthewferry's comment here "this doesn't account for active states of buttons not using popovers or all button styles including minimal and primary buttons."  is my primary concern. It conflates the Button component with additional uncommon use cases. What additional value does the "active" color on a Button with open Popover add for users?

PR #1667, opened 2025-05-12
fix: fix long headers overflowing on modal

tracisiebel:
## Summary

Screenshot 2025-05-12 at 4 55 21 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -104,5 +104,10 @@
 			"vite": "npm:rolldown-vite@6.3.6"
 		}
 	},
-	"browserslist": ["last 2 versions", "not dead", "not IE 11"]
+	"browserslist": ["last 2 versions", "not dead", "not IE 11"],
+	"dependencies": {
+		"@launchpad-ui/components": "0.9.22",
+		"@launchpad-ui/icons": "0.21.11",
+		"@launchpad-ui/tokens": "0.12.4"
+	}
 }
--- a/packages/components/src/styles/Heading.module.css
+++ b/packages/components/src/styles/Heading.module.css
@@ -5,19 +5,3 @@
 	text-overflow: ellipsis;
 	position: relative;
 }
-
-.heading:hover {
-	cursor: default;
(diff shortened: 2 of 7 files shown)


package.json
```diff
@@ -104,5 +104,10 @@
 			"vite": "npm:rolldown-vite@6.3.6"
 		}
 	},
-	"browserslist": ["last 2 versions", "not dead", "not IE 11"]
+	"browserslist": ["last 2 versions", "not dead", "not IE 11"],
+	"dependencies": {
+		"@launchpad-ui/components": "0.9.22",
+		"@launchpad-ui/icons": "0.21.11",
+		"@launchpad-ui/tokens": "0.12.4"
+	}
```
apucacao, line 112: Do we need these changes?

packages/components/src/styles/Heading.module.css
```diff
@@ -1,2 +1,7 @@
 .heading {
+	white-space: nowrap;
+	max-width: calc(100% - var(--lp-size-32));
```
apucacao, line 3: There's a small downside to this `max-width`: what if the modal is _not_ dismissible and doesn't have the button? We're not going to make use of the 32px, which seems small, but could contain literally 2-3 letters. And with this rule they'd be truncated. It's a bit of an edge case for sure, but still.

I was playing around with the layout in DevTools and it seemed a bit strange to me that we rely on `flex-wrap: wrap` to ensure the subtitle appears below the heading. Seems like using a grid could help simplify things a bit and avoid that `max-width`.

We can rethink that after this quickfix though. This at least will deal with super long titles better.

apucacao (approved):

package.json
apucacao: @traciklopez i forgot about this comment -- this should not have gone in, unless you had a good reason?

traciklopez: oh shoot. I thought this was from the changelog command.

PR #1669, opened 2025-05-14
feat(components): Set select and combobox popover container width to be the width of the trigger

cmaija:
## Summary

This PR updates the popover so that it has a `width` of its trigger, which was the [result of a conversation in slack](https://launchdarkly.slack.com/archives/CDXEFNMLP/p1747168233434969)

before:
Screenshot 2025-05-14 at 2 10 53 PM

after:
Screenshot 2025-05-14 at 2 10 32 PM

## Testing approaches

- visual check in storybook

### DIFF
--- a/.changeset/soft-eggs-wave.md
+++ b/.changeset/soft-eggs-wave.md
@@ -1,0 +1,5 @@
+---
+"@launchpad-ui/components": patch
+---
+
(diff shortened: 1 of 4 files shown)


apucacao (approved):
Thanks @cmaija!

apucacao:
I'm going to merge this because there's another change we want to make sure goes out too.

PR #1703, opened 2025-06-03
fix: Fix i8n errors by adding spans to button children

yusinto:
## Summary

https://launchdarkly.atlassian.net/browse/SC-295319

The gonfalon-fe channel receives many i8n errors like so:

```sh
ReactRenderingError: Failed to execute 'insertBefore' on 'Node': The node before which the new node is to be inserted is not a child of this node.

ReactRenderingError: The object can not be found here.
```

This is related to how the react dom gets removed/added by google translate on chrome. Read more about it [here](https://martijnhols.nl/blog/everything-about-google-translate-crashing-react).

---
Related Jira issue: [SC-295319: React i8n dom errors](https://launchdarkly.atlassian.net/browse/SC-295319)

### DIFF
--- a/.changeset/sixty-steaks-jump.md
+++ b/.changeset/sixty-steaks-jump.md
@@ -1,0 +1,5 @@
+---
+"@launchpad-ui/components": minor
+---
+
+Add spans to button children to fix i8n errors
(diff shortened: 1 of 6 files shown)


nhironaka (approved):

Niznikr (changes requested):
Several visual regressions in Chromatic. I wonder if this is a worthwhile change and worth the tradeoff (more dom nodes) given it doesn't fully solve the issue as pointed out by the link you shared? That and we would have to update all the other components with similar structure as I doubt button is the only culprit.

PR #1725, opened 2025-07-24
feat(components): Typography token styles added to Heading, Text, Label, Code components

Zuzze:
## Summary

To streamline text usage on instance level, `Text`, `Label` and `Heading` support now the following optional props (No breaking changes):

- `variant`: if not given, uses old default style as before
- `maxLines`: default off like before

New shorthand syntax looks like

- `<Text>Lorem ipsum dolor sit est</Text>` -> defaults to `<Text size="medium" bold={false}>`
- `<Heading>Welcome</Heading>` ->  defaults to`<Heading size="medium" bold={true}>`
- `<Label>Name</Label>` -> defaults to`<Label size="medium">`
- `<Code>const var = 1</Code>` -> defaults to `\<Code size="medium">`

### New props for `Heading`,  `Text`, `Code`, `Label`:
- `bold`: If text is emphasized, will be true. Design team will align so that Figma variants will only have 2 font weights for each. Label has only one font weight in the props API but if inside form where orientation="horizontal" lighter weight will be applied.
- `size`: `small`, `medium`, `large` will be used. Label and Code have only 2 sizes. More sizes may be added in future

Under the hood the component also handles:
- Semantic HTML tag (e.g. `code` for code, `small` for small text and `p` for bigger text and levels for headings
- mapping to corresponding design system tokens in css modules

Underlying HTML token can still be changed by react-aria props' `elementType` in Text and `level` in Heading like before. 

### Text

Screenshot 2025-07-30 at 11 08 19

Screenshot 2025-07-30 at 11 08 30

### Heading

Screenshot 2025-07-30 at 11 27 23

Screenshot 2025-07-30 at 11 27 29

### Label

image

### Code

Screenshot 2025-07-30 at 09 40 40

### Housekeeping

- Added links to new `Text` and `Heading`, `Code`, `Label` from `Typography` tokens
- Added storybook config that displays types inside union instead of `union` in prop table

### Dark mode

All text components were also tetsted in dark mode.

image

## Testing approaches

New stories will be detected in future by Chromatic

### DIFF
--- a/packages/components/src/Button.tsx
+++ b/packages/components/src/Button.tsx
@@ -72,7 +72,11 @@ const Button = ({ ref, ...props }: ButtonProps) => {
 					{isPending && (
 						<ProgressBar isIndeterminate aria-label="loading" className={styles.progress} />
 					)}
-					{typeof children === 'string' ? <Text>{children}</Text> : children}
+					{typeof children === 'string' ? (
+						<Text size={size as 'small' | 'medium' | 'large'}>{children}</Text>
+					) : (
+						children
+					)}
 				</Provider>
 			))}
 		</AriaButton>
--- a/packages/components/src/Heading.tsx
+++ b/packages/components/src/Heading.tsx
@@ -8,82 +8,75 @@ import { Heading as AriaHeading } from 'react-aria-components';
 import styles from './styles/Heading.module.css';
 import { useLPContextProps } from './utils';
 
-/** variant names are based on Figma's naming convention */
-export type HeadingVariant =
-	| 'display1'
-	| 'heading1Medium'
-	| 'heading1Semibold'
-	| 'heading1ExtraBold'
-	| 'heading2Medium'
-	| 'heading2Semibold'
-	| 'heading2ExtraBold'
-	| 'heading3Regular'
-	| 'heading3Semibold'
-	| 'heading3ExtraBold';
-
 const headingStyles = cva(styles.heading, {
 	variants: {
-		variant: {
-			display1: styles.display1,
-			heading1Medium: styles.heading1Medium,
-			heading1Semibold: styles.heading1Semibold,
-			heading1ExtraBold: styles.heading1ExtraBold,
-			heading2Medium: styles.heading2Medium,
-			heading2Semibold: styles.heading2Semibold,
-			heading2ExtraBold: styles.heading2ExtraBold,
-			heading3Regular: styles.heading3Regular,
-			heading3Semibold: styles.heading3Semibold,
-			heading3ExtraBold: styles.heading3ExtraBold,
+		size: {
+			small: styles.small,
+			medium: styles.medium,
+			large: styles.large,
 		},
+		bold: {
+			true: styles.bold,
+		},
+	},
+	defaultVariants: {
+		size: 'medium',
+		bold: true,
 	},
 });
 
 interface HeadingProps extends Omit<AriaHeadingProps, 'className' | 'level'> {
 	ref?: Ref<HTMLHeadingElement>;
-	/** Typography variant following LaunchPad design system typography tokens. */
-	variant?: HeadingVariant;
+	/** Heading size */
+	size?: 'small' | 'medium' | 'large';
+	/** Whether to use bold font weight */
+	bold?: boolean;
 	/** Maximum number of lines to display. Overflowing text will be truncated with an ellipsis. */
 	maxLines?: number;
 	/** Optional CSS class name */
 	className?: AriaHeadingProps['className'];
-	/** Optional HTML heading level. Defaults to variant level. */
+	/** Optional HTML heading level. Defaults based on size (large=1, medium=2, small=3). */
 	level?: AriaHeadingProps['level'];
 }
 
 const HeadingContext = createContext<ContextValue<HeadingProps, HTMLHeadingElement>>(null);
 
-const getDefaultLevel = (variant: HeadingVariant): AriaHeadingProps['level'] => {
-	switch (variant) {
-		case 'display1':
-		case 'heading1Medium':
-		case 'heading1Semibold':
-		case 'heading1ExtraBold':
+const getDefaultLevel = (size: 'small' | 'medium' | 'large'): AriaHeadingProps['level'] => {
+	switch (size) {
+		case 'large':
 			return 1;
-		case 'heading2Medium':
-		case 'heading2Semibold':
-		case 'heading2ExtraBold':
+		case 'medium':
 			return 2;
-		default:
+		case 'small':
 			return 3;
 	}
 };
 
 /**
- * A generic Heading component for headings and display text.
+ * A generic Heading component for headings.
  *
- * For body text, use [Text](/docs/components-content-text--docs)
+ * For other text, use `Text`, `Label`, or `Code`.
  *
  * Built on top of [React Aria `Heading` component](https://react-spectrum.adobe.com/react-spectrum/Heading.html#heading).
  */
-const Heading = ({ ref, variant, maxLines, className, style, level, ...props }: HeadingProps) => {
+const Heading = ({
+	ref,
+	size = 'medium',
+	bold = true,
+	maxLines,
+	className,
+	style,
+	level,
+	...props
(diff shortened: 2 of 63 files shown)


packages/components/src/Text.tsx
```diff
@@ -1,26 +1,126 @@
...
+	/** Typography variant following LaunchPad design system Typography tokens. */
+	variant?: TextVariant;
+	/** Text alignment */
+	align?: 'left' | 'center' | 'right';
+	/** Maximum number of lines to display. Overflowing text will be truncated with an ellipsis. */
+	maxLines?: number;
+	/** Optional HTML element type such as `span`, `label`, `small`, `code`, `p`, etc. Defaults to the appropriate semantic element based on the variant. */
+	elementType?: AriaTextProps['elementType'];
+	/** Optional CSS class name */
+	className?: AriaTextProps['className'];
```
Zuzze, line 71: In case you're wondering: The reason why these are extracted here and omitted on top is to add custom TSDoc

1) for devs so they can read more about this directly via Intellisense in the IDE when hovering:

image

2) Additionally, same comment will be automatically displayed in the stories.tsx and avoiding people to need to jump into React-aria docs for the most common props:

image

packages/components/src/Text.tsx
```diff
@@ -1,26 +1,126 @@
...
+			label2Regular: styles.label2Regular,
+			label2Medium: styles.label2Medium,
+			code1Regular: styles.code1Regular,
+			code2Regular: styles.code2Regular,
+		},
+		align: {
+			left: styles.left,
+			center: styles.center,
+			right: styles.right,
+		},
```
apucacao, line 52: I'm less sure about these mostly because these have nothing to do with our design system. They're mostly here for convenience, right?

apucacao: This just reminded me: Robb (previous LP ENG lead) and I had talked about what it might look like to have a Tailwind theme that mapped to our design system, so that all tokens etc were a class name away.

I have opinions about Tailwind, but, the desire from devs on the team to be able to adhere to our design system _and_ assemble UIs quickly keeps coming up, and I'm feeling more enticed by it myself despite my reservation primarily because I think it might make agents' lives easier too!

Is that something you've done before? Curious if you have thoughts around this!

apucacao: Oh sorry forgot to answer your question!

I think it's OK to keep `align` for now. And let's also think about potentially fleshing this out further into a fuller system!

apucacao (approved):

packages/components/src/Text.tsx
apucacao: I know @nhironaka has thoughts about this.

apucacao: > Let's see what others think and we can add that in a separate MR, ok?
👍 

Yeah, I think there's maybe a hierarchy here:
- Crawl phase / early iteration: Tailwind or similar
- Walk phase and beyond: Components (that enforce tighter constraints if needed)

Introducing abstractions early on feels risky to me, and also may add friction which could interfere with quick iteration / play.

In any case, this looks 👍 to me. We can have a bigger convo about the rest!

This would be a good thing to share in #eng-frontend once you've bumped the Gonfalon version.

nhironaka (approved):
nice!

packages/components/src/Text.tsx
Zuzze: Sounds good, will share on the channel after npm release! 👍

matthewferry (changes requested):
This is dope! And thank you for proactively getting this done! Just a couple things:

1. let's remove labels from being used in the api for `<Text/>`. we have `<Label />` for that and don't want to encourage those styles in regular copy
2. I find the use of single variant prop a bit awkward. I can be peaceful with us doing it for this pass (I know there are feels about expanding props), but I would prefer the API be explicit for `size` and `weight`.

packages/components/src/Heading.tsx
```diff
@@ -1,26 +1,99 @@
...
+			display1: styles.display1,
+			heading1Medium: styles.heading1Medium,
+			heading1Semibold: styles.heading1Semibold,
+			heading1ExtraBold: styles.heading1ExtraBold,
+			heading2Medium: styles.heading2Medium,
+			heading2Semibold: styles.heading2Semibold,
+			heading2ExtraBold: styles.heading2ExtraBold,
+			heading3Regular: styles.heading3Regular,
+			heading3Semibold: styles.heading3Semibold,
+			heading3ExtraBold: styles.heading3ExtraBold,
```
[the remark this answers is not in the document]
Zuzze: Thanks for your feedback @matthewferry! The reason why these were coupled into one prop instead of separate props is that having them in separate props would end up in "impossible" combinations that do not exist in Figma which could cause deviation from the "officially approved" variations.

Currently in Figma, we have 1-3 variations on each type of text so we would need to throw an error for "empty" instances like `<Heading size="1" weight="regular">` or <Text size="2" weight="medium". Also the `size` naming may be misleading as for instance we have also `display` variant which can be used from SEO perspective as `h1` but in other pages `heading1` may be the semantic `h1`. 

If weight is wanted to keep as a separate prop, another pattern I have seen would be to introduce own components for each  `<Code>`, `<Body1>`, `<Body2>`, `<Heading1>` etc so that there would not be room for these "impossible" states.

image

packages/components/src/Text.tsx
```diff
@@ -1,26 +1,106 @@
...
+	| 'body2ExtraBold'
+	| 'small1Regular'
+	| 'small1Medium'
+	| 'small1Semibold'
+	| 'label1Regular'
+	| 'label1Medium'
+	| 'label2Regular'
+	| 'label2Medium'
+	| 'code1Regular'
+	| 'code2Regular';
```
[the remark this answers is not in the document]
Zuzze: Figma is used as a single source of truth here and the goal is to have 1:1 mapping between the Figma and code variants to make the implementation as efficient as possible. 

Happy to change the naming convention here but then would be good to align that same convention also on Figma design system, otherwise we may get lost in translation.

apucacao:
> bold: If text is emphasized

Just curious: would it make sense to use "emphasized" or something similar to make clear the prop semantics vs visual effect?

apucacao (approved):
Nice work driving this, @Zuzze!

Zuzze:
> > bold: If text is emphasized
> 
> Just curious: would it make sense to use "emphasized" or something similar to make clear the prop semantics vs visual effect?

Could be, `bold` was just shorter and less prone to typos I guess 😅 But if others also prefer `emphasized` or have some better word in mind, I'm not against that either!

matthewferry (approved):
LGTM!

packages/components/src/Button.tsx
```diff
@@ -72,7 +72,11 @@ const Button = ({ ref, ...props }: ButtonProps) => {
 					{isPending && (
 						<ProgressBar isIndeterminate aria-label="loading" className={styles.progress} />
 					)}
-					{typeof children === 'string' ? <Text>{children}</Text> : children}
+					{typeof children === 'string' ? (
+						<Text size={size as 'small' | 'medium' | 'large'}>{children}</Text>
```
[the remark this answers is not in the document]
Zuzze: Yeah I was also confused by this, existing storybook story examples used nested text inside button which I did not exactly follow (image below). 

This fix inside the button avoided breaking changes based on chromatic snapshot tests so devs won't be able to change only text size, it will be defined via button size only (unless they add custom ReactNodes as prop).

image

packages/form/stories/Label.stories.tsx
```diff
@@ -5,9 +5,16 @@ import { Label } from '../src';
...
 	component: Label,
 	title: 'Legacy/Form/Label',
-	description: 'A label describes a form field.',
+
 	parameters: {
 		chromatic: { disableSnapshot: true },
+		docs: {
+			description: {
+				component: `
+A label describes a form field. This component is deprecated. Use [Label](/docs/components-content-label--docs) from \`@launchpad-ui/components\` instead.
```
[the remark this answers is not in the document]
Zuzze: Sounds like a good idea! 👍

PR #1738, opened 2025-08-08
feat: Add component attribution and Afterburn developer tool for identifying in-use components

zmdavis:
Screenshot 2025-08-08 at 7 18 50 AM


nhironaka (approved):
nice

since launchpad is a public repo, are there any security concerns for us exposing this?

zmdavis:
> nice
> 
> since launchpad is a public repo, are there any security concerns for us exposing this?

I can't think of any? It's an opt-in and doesn't expose anything that isn't already publicly available.

packages/attribution/dist/index.es.js
```diff
@@ -0,0 +1,12 @@
+/**
```
Niznikr, line 1: Is checking in these `dist` files intentional?

PR #1748, opened 2025-08-25
fix(components): removal buttons overflow small Tags

pheggeseth:
## Summary
Tags of size `"small"` that were also removable had icon buttons that were still 24 x 24, so they would overflow the tag container on hover or focus.

Before:

https://github.com/user-attachments/assets/55a74b51-f723-4831-aefd-1fb60269f609

After:

https://github.com/user-attachments/assets/41b0042c-accb-49da-8ac7-177ce3aaf891


nhironaka (approved):
ty!

PR #1750, opened 2025-08-25
fix(components): set min-width on Alert content to let content shrink to fit container

pheggeseth:
## Summary
Sets a `min-width: 0` on the `.content` wrapper inside `Alert`, which lets the content wrapper shrink to fit inside the root flex container. This typically only becomes an issue if the Alert itself has a constrained width and if some content within the Alert forces the `.content` wrapper to grow, such as if the content contains a `<div>` and that `<div>` contains a `<span>` with a very large word in it.

Before:
Screenshot 2025-08-25 at 3 47 34 PM

After:
Screenshot 2025-08-25 at 3 53 12 PM

With this change, there's still the problem of the overflowing content, but because the `.content` wrapper has a calculable width and shrinks to respect its flex container, consumers can properly style the slotted children as they see fit, such as setting a `max-width: 100%` and having that width be calculated property based on the width of the parent.

### DIFF
--- a/.changeset/chatty-insects-clean.md
+++ b/.changeset/chatty-insects-clean.md
@@ -1,0 +1,5 @@
+---
+"@launchpad-ui/components": patch
+---
+
+Set min-width on Alert content wrapper to let it shrink to fit flex container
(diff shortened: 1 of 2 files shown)


apucacao (approved):
classic!

PR #1759, opened 2025-08-28
fix(components): remove duplicate Tag name from Tag removal IconButton accessible name

pheggeseth:
## Summary
In testing the usage of a `Tag`'s removal button when a `TagGroup` has a `onRemove` callback, I realized that by including the `Tag`'s `textValue` as part of the removal `IconButton`'s `aria-label`, this results in a final calculated accessible name for that button of `Remove tag-name tag-name` instead of just `Remove tag-name`, so querying for the button in unit tests a la `screen.getByRole('button', { name: 'Remove tag-1' })` wouldn't work unless you repeated the tag name like: `screen.getByRole('button', { name: 'Remove tag-1 tag-1' })`.

## Testing approaches
Updated tests so that we're asserting that the accessible name is what we expect, rather than just what the `aria-label` of the removal button is.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -7,5 +7,8 @@
 		"quickfix.biome": "explicit",
 		"source.organizeImports.biome": "explicit"
 	},
-	"editor.tabSize": 2
+	"editor.tabSize": 2,
+	"[typescriptreact]": {
+		"editor.defaultFormatter": "biomejs.biome"
+	}
 }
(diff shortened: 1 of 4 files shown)


nhironaka (approved):

.vscode/settings.json
```diff
@@ -7,5 +7,8 @@
 		"quickfix.biome": "explicit",
 		"source.organizeImports.biome": "explicit"
 	},
-	"editor.tabSize": 2
+	"editor.tabSize": 2,
+	"[typescriptreact]": {
+		"editor.defaultFormatter": "biomejs.biome"
+	}
```
pheggeseth, line 13: FYI at least for me in Cursor, it was trying to use Prettier to format tsx files, so I went ahead and added this.

PR #1795, opened 2025-10-20
feat(components): move the snippet and copy to code components from gonfalon to LP

ari-launchdarkly:
## Summary

- Ports over the CopyToClipboard component and utility methods from gonfalon to LP
- Ports over the Snippet component and styling to LP
- Adds the requisite dependencies as part of porting over the components

https://launchdarkly.atlassian.net/browse/REL-10213

Screenshot 2025-10-20 at 13 13 03

## Testing approaches

---
Related Jira issue: [REL-10213: Move Snippet and CopyToClipboard to launchpad-ui](https://launchdarkly.atlassian.net/browse/REL-10213)

### DIFF
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -33,7 +33,10 @@
 		"@internationalized/date": "3.10.0",
 		"@launchpad-ui/icons": "workspace:~",
 		"@launchpad-ui/tokens": "workspace:~",
-		"class-variance-authority": "0.7.0"
+		"@react-aria/live-announcer": "3.4.4",
+		"class-variance-authority": "0.7.0",
+		"prism-themes": "1.9.0",
+		"prismjs": "1.30.0"
 	},
 	"devDependencies": {
 		"@react-aria/focus": "3.21.2",
@@ -41,6 +44,7 @@
 		"@react-aria/utils": "3.31.0",
 		"@react-stately/utils": "3.10.8",
 		"@react-types/shared": "3.32.1",
+		"@types/prismjs": "1.26.5",
 		"copyfiles": "2.4.1",
 		"react": "19.2.0",
 		"react-aria": "3.44.0",
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -35,8 +35,7 @@
 		"@launchpad-ui/tokens": "workspace:~",
 		"@react-aria/live-announcer": "3.4.4",
 		"class-variance-authority": "0.7.0",
-		"prism-themes": "1.9.0",
-		"prismjs": "1.30.0"
+		"prism-themes": "1.9.0"
 	},
 	"devDependencies": {
 		"@react-aria/focus": "3.21.2",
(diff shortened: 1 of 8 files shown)


packages/components/src/utils.tsx
```diff
@@ -91,4 +93,69 @@ const useLPContextProps = <T, U extends SlotProps, E>(
...
+			title:
+				toastMessage ??
+				`'${text.length > MAX_WIDTH ? `${text.slice(0, MAX_WIDTH)}...` : text}' copied to clipboard.`,
+			status: 'success',
+		});
+		announce('Copied!', 'polite', 300);
+		return true;
+	} catch (error) {
+		announce('Failed to copy', 'polite', 300);
+		console.error(`Unable to copy: ${error}`);
```
ari-launchdarkly: I'm not sure what to do about this, we were previously sending up a `trackError` and throwing this error message. Is there a pattern we want to establish where we can throw a CustomEvent that the consumer listens for if there's a failure?

yusinto: I think the previous `trackError` solution wasn't very useful anyway so I think we can safely just `announce` and remove the console.error and not do anything else.

packages/components/package.json
```diff
@@ -33,14 +33,18 @@
 		"@internationalized/date": "3.10.0",
 		"@launchpad-ui/icons": "workspace:~",
 		"@launchpad-ui/tokens": "workspace:~",
-		"class-variance-authority": "0.7.0"
+		"@react-aria/live-announcer": "3.4.4",
```
yusinto: Should this be in devDep? I see all the other `@react-aria` deps are devDeps. I'm not sure why, but best to stay consistent. Better still find out why they are devDep and not dep.

yusinto (approved):
Couple of comments, but otherwise lgtm.

packages/components/package.json
ari-launchdarkly: I'm not sure on that one. The reason being that we had originally had it as a dependency in the dependencies list and I didn't want to stray from that. What'd you think?

https://github.com/launchdarkly/gonfalon/blob/main/packages/forms-experimental/package.json#L6-L28

apucacao:
Is this related to us trying to share onboarding code between different code bases?

packages/components/package.json
```diff
@@ -33,14 +33,18 @@
...
+		"class-variance-authority": "0.7.0",
+		"prism-themes": "1.9.0",
+		"prismjs": "1.30.0"
```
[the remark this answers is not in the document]
ari-launchdarkly: @yusinto and I talked about it. I'm not sure if we want is a peer dependency, or as a direct dependency. The [package size ](https://bundlephobia.com/package/prismjs@1.30.0) for a minified bundle is at 18.7K (minified and gzipped at 6.9K)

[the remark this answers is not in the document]
apucacao: i feel like a peer dependency would work well here? Curious what @vezaynk thinks too.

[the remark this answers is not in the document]
yusinto: Snippet should be a share-able LP component. It's been living in launchpad-experimental for 3 years or so with that prismjs dependency. I did look at the size yes ari is correct. That dep must exist somewhere.

Screenshot 2025-10-22 at 11 10 18 AM

apucacao (approved):

PR #1808, opened 2025-10-30
feat(components): add `toastMessage` prop to `CopyToClipboard`

Andrewjeska:
## Summary

I'm adding a prop to `CopyToClipboard` so the consumer can customize the toast message that's show on copy

## Testing approaches

I didn't see a unit test for this component, should there be?


ari-launchdarkly (approved):
Thanks for this. It's gonna need approval from the FM team

nhironaka (approved):

PR #1829, opened 2025-12-16
fix(components): addresses the modal closing when toasts appear bug

ari-launchdarkly:
## Summary

- Adds a test case to validate against how dialogs behave with toasts. 
- Updates the Modal component to prevent spamming buttons that trigger a toast from closing the dialog. 

### Problem

The crux of the problem here is that react-aria uses a portal mechanism to set toasts onto when they are appended onto the DOM. When someone triggers a toast, a HTML tag is appended to the DOM via the portal, so spamming the same button results in the second click (after the toast is initialized) from being set onto the HTML tag, which is "outside" the dialog and causes it to collapse. This solution works by listening to the element that's being clicked on and preventing the collapse when that's a BODY or HTML tag.

## CURRENT BUG:

https://www.loom.com/share/98745d81ab324abeb91cdf522b5c47c5

## SOLUTION:

https://www.loom.com/share/6a0880c893184250a6a0b671eb4d5a66

## Testing approaches

---
Related Jira issue: [REL-11286: BUG: Clicking a dialog when a toast is up closes the dialog](https://launchdarkly.atlassian.net/browse/REL-11286)

### DIFF
--- a/packages/components/src/Modal.tsx
+++ b/packages/components/src/Modal.tsx
@@ -94,6 +94,19 @@ const ModalOverlay = ({ isDismissable = true, ref, ...props }: ModalOverlayProps
 	return (
 		<AriaModalOverlay
 			isDismissable={isDismissable}
+			shouldCloseOnInteractOutside={(element) => {
+				// Don't close when the element being checked is HTML or BODY.
+				// This indicates a portal interaction (e.g., spawning a toast from a button).
+				//
+				// Why this is safe:
+				// - When users click the modal backdrop, the element is the overlay <div>, NOT HTML/BODY
+				// - HTML/BODY only appear here during React portal rendering edge cases
+				// - This prevents the modal from incorrectly closing when triggering toasts from modal buttons
+				// - Legitimate outside clicks (backdrop, other UI) still close the modal as expected
+				const isRootElement = element.tagName === 'HTML' || element.tagName === 'BODY';
+
+				return !isRootElement;
+			}}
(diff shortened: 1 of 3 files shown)


packages/components/src/Modal.tsx
```diff
@@ -94,6 +94,19 @@ const ModalOverlay = ({ isDismissable = true, ref, ...props }: ModalOverlayProps
 	return (
 		<AriaModalOverlay
 			isDismissable={isDismissable}
+			shouldCloseOnInteractOutside={(element) => {
```
nhironaka: @cspath1 will this negatively interact with toolbar?

Also curious if this will have a negative impact on public consumers (who may be rendering any number of iframes etc)

cspath1: I don't think this change will negatively interact with the toolbar thankfully! As far as Modal usage within the toolbar is concerned, we aren't using the modal component (we have a popup window for the auth flow, since storing cookies via an iframe can cause issues).

More generally speaking though, I do think there is an _existing_ issue with any sort of auto-close functionality when it comes to interacting with the toolbar (i.e. you wouldn't be able to keep the modal open and also interact with the toolbar without the modal closing). This isn't too big of a deal, unless you want to use the toolbar specifically to update flag values/track events that occur within said modal.

I'm curious if there's anything we (FM Next) can do to mitigate that sort of issue on our end.

nhironaka (approved):

PR #1931, opened 2026-06-11
fix(tokens): show gradient value and hex in color token docs

hsadhvani:
## Summary

The Tokens/Color Storybook docs read getComputedStyle().backgroundColor, which is rgba(0, 0, 0, 0) for gradient tokens since they are painted via background-image. Read background-image for gradients and display the gradient definition instead of the meaningless transparent value.

Also derive and show a hex value for solid color tokens, keeping the rgba string alongside it for backwards compatibility.

CleanShot 2026-06-11 at 16 05 06

## Testing approaches
Check launchpad storybook as it's only a story change 

---

> [!NOTE]
> **Low Risk**
> Documentation-only Storybook and story-module changes with no runtime product impact.
> 
> **Overview**
> Fixes the **Tokens/Color** Storybook table so gradient tokens show their real **`background-image`** value instead of misleading transparent **`backgroundColor`**, and adds a **Hex** column for solid tokens while keeping **rgba** for compatibility.
> 
> Display logic moves into **`colorTokens`** (`rgbToHex`, `toRgba`, `getTokenHex`, `getTokenValue`, **`TokenCode`**), with **`TokenTable`** now storing computed **`{ color, image }`** per swatch. Gradients show **—** in Hex and the gradient string in rgba; solids show uppercase hex plus normalized rgba (muted when hex is primary). **`colorTokens.spec.tsx`** covers the helpers and **`TokenCode`** rendering.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 18d473c8b9b54e42fa3e78a34b2222d0afc2ef8e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


zachalbert (approved):
Thanks for doing this! Just a minor note, but it looks like we're still showing `rgb` rather than `rgba`. I don't know if we're really using rgba anywhere anyway so it probably doesn't matter, but just in case.

hsadhvani:
> Thanks for doing this! Just a minor note, but it looks like we're still showing `rgb` rather than `rgba`. I don't know if we're really using rgba anywhere anyway so it probably doesn't matter, but just in case.

Good catch, just fixed it. Also, I don't have the authority to merge here, so I leave that to you :)

PR #1939, opened 2026-06-23
fix(components): prevent overlay subtitle slot from leaking into nested Alert

ShreyasGit51283:
## Summary

Fixes [DSYS-157](https://launchdarkly.atlassian.net/browse/DSYS-157). When an `Alert` is nested inside an overlay (`Dialog`/`Modal`), the overlay's `subtitle` slot context leaked into the Alert's `Text`:

- It **forced** `slot="subtitle"` on the Alert's `Text` (a bare `<Text>` threw `"A slot prop is required..."`).
- The text inherited the Dialog's `id` + `elementType: 'h3'`, so the **Alert body became the Dialog's `aria-describedby` description rendered as an `<h3>`** (an accessibility bug).
- The text picked up the overlay's `[slot="subtitle"]` CSS (wrong color/size/position).

### Root cause
- `Dialog` provides a RAC `TextContext` with `slots: { subtitle: { id, elementType: 'h3' } }`.
- `Alert` only overrode `HeadingContext`/`ButtonGroupContext`, so its descendants inherited the Dialog's `TextContext`.
- `Modal.module.css` styled `[slot="subtitle"]` via a **descendant** selector, reaching into nested alerts.

### Changes
- **`Alert.tsx`** — establish Alert's own `TextContext` (default slot + `subtitle` slot). Insulates Alert content from ancestor overlays, fixes the a11y leak, and allows a bare `<Text>` inside an Alert. Backward-compatible: existing `<Text slot="subtitle">` usage still works.
- **`Modal.module.css`** — scope the subtitle rule with `:not([role='alert'] *)` so it no longer reaches into nested alerts. The overlay's own subtitle is unaffected.
- **Tests** — `Alert.spec.tsx` adds nested-in-Dialog regression coverage (verified to fail without the fix with the exact `"A slot prop is required"` error).
- **Story** — `Alert.stories.tsx` adds "Nested in Dialog (DSYS-157)" for visual + Chromatic regression coverage.

## Test plan
- [ ] `Components/Status/Alert` -> "Nested in Dialog (DSYS-157)": both nested alerts render with Alert text styling; the Dialog's own subtitle keeps overlay styling.
- [ ] Existing Alert/Dialog/Modal stories and tests unaffected (`pnpm --filter @launchpad-ui/components test`).
- [ ] Confirm a bare `<Text>` inside an Alert (in and out of a Dialog) renders without throwing.
- [ ] Chromatic diff review for Alert/Dialog/Modal.

[DSYS-157]: https://launchdarkly.atlassian.net/browse/DSYS-157?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/components/src/Alert.tsx
+++ b/packages/components/src/Alert.tsx
@@ -5,7 +5,8 @@ import { StatusIcon } from '@launchpad-ui/icons';
 import { useControlledState } from '@react-stately/utils';
 import { cva } from 'class-variance-authority';
 import { HeadingContext } from 'react-aria-components/Heading';
-import { Provider } from 'react-aria-components/slots';
+import { DEFAULT_SLOT, Provider } from 'react-aria-components/slots';
+import { TextContext } from 'react-aria-components/Text';
 
 import { ButtonGroupContext } from './ButtonGroup';
 import { IconButton } from './IconButton';
@@ -99,6 +100,19 @@ const Alert = ({
 					values={[
 						[HeadingContext, { className: styles.heading }],
 						[ButtonGroupContext, { className: styles.buttonGroup }],
+						// Establish Alert's own Text context so an ancestor overlay (e.g. a Dialog,
+						// which sets a `subtitle` slot with its own id/elementType/CSS) cannot leak
+						// into Alert content. Provides a default slot so a bare <Text> works, plus a
+						// `subtitle` slot for backward compatibility with existing usage.
+						[
+							TextContext,
+							{
+								slots: {
+									[DEFAULT_SLOT]: {},
+									subtitle: {},
+								},
+							},
+						],
 					]}
 				>
 					{children}
--- a/packages/components/src/Alert.tsx
+++ b/packages/components/src/Alert.tsx
@@ -100,10 +100,6 @@ const Alert = ({
 					values={[
 						[HeadingContext, { className: styles.heading }],
 						[ButtonGroupContext, { className: styles.buttonGroup }],
-						// Establish Alert's own Text context so an ancestor overlay (e.g. a Dialog,
-						// which sets a `subtitle` slot with its own id/elementType/CSS) cannot leak
-						// into Alert content. Provides a default slot so a bare <Text> works, plus a
-						// `subtitle` slot for backward compatibility with existing usage.
 						[
 							TextContext,
 							{
--- a/packages/components/src/styles/Modal.module.css
+++ b/packages/components/src/styles/Modal.module.css
@@ -84,7 +84,10 @@
 		gap: var(--lp-size-10);
 	}
 
-	& [slot='subtitle'] {
+	/* Scope to the overlay's own subtitle. The `:not()` keeps this from reaching into a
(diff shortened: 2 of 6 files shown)


packages/components/src/Alert.tsx
```diff
@@ -99,6 +100,19 @@ const Alert = ({
 					values={[
 						[HeadingContext, { className: styles.heading }],
 						[ButtonGroupContext, { className: styles.buttonGroup }],
+						// Establish Alert's own Text context so an ancestor overlay (e.g. a Dialog,
+						// which sets a `subtitle` slot with its own id/elementType/CSS) cannot leak
+						// into Alert content. Provides a default slot so a bare <Text> works, plus a
+						// `subtitle` slot for backward compatibility with existing usage.
```
nhironaka: we can remove this comment

packages/components/src/styles/Modal.module.css
```diff
@@ -84,7 +84,10 @@
 		gap: var(--lp-size-10);
 	}
 
-	& [slot='subtitle'] {
+	/* Scope to the overlay's own subtitle. The `:not()` keeps this from reaching into a
```
nhironaka: mentioned this during the meeting, but let's undo this change for now

packages/components/stories/Alert.stories.tsx
```diff
@@ -236,3 +240,56 @@ export const InlineTones: Story = {
...
+// Nested overlay regression (DSYS-157)
+// =============================================================================
+
+/**
+ * Regression coverage for DSYS-157. When Alerts are nested inside a Dialog/Modal, the
+ * overlay's `subtitle` slot context (styling + semantics) must not leak into the Alert's
+ * text. The Dialog's own subtitle keeps the overlay styling, while both nested Alerts
+ * render with the Alert's body-text styling — identical to how they look outside a modal.
+ */
+export const NestedInDialog: Story = {
```
zakk-verrilli-ld: I moved this to the Modal stories as a "With header" story because I realized we haven't documented the `header` slot at all. Also removed the ticket from the story name so we don't conflate LP with internal data

zakk-verrilli-ld (approved):

packages/components/src/styles/Modal.module.css
zakk-verrilli-ld: I see where you're coming from @nhironaka. I refactored the selectors ([commit](https://github.com/launchdarkly/launchpad-ui/pull/1939/commits/a91b5a1b9622d294cbc401607b3eef30f33de865)) to make it explicit that we support `slot="header"` as a direct child of Modal as well as nested in Dialog, but I made a [ticket](https://launchdarkly.atlassian.net/browse/DSYS-216) to deprecate the former, as I see a few usages in Gonfalon

nhironaka: sounds good!

zakk-verrilli-ld (approved):

PR #1941, opened 2026-06-25
feat(components): [DSYS-194] add description and errorMessage props to field components

ShreyasGit51283:
## Summary

Resolves the `description` prop drift between the LaunchPad React components and the Figma library ([DSYS-194](https://launchdarkly.atlassian.net/browse/DSYS-194)).

The Figma library exposes a flat `description` property on field components, but in code helper text was only expressible as a slotted child (`<Text slot="description">`). This mismatch broke Code Connect mappings.

This PR adds ergonomic `description` and `errorMessage` props to the field wrappers, implemented through a single shared helper so the slot-rendering logic lives in exactly one place.

- New `renderFieldHelpers()` + `FieldHelperProps` in `src/fieldHelpers.tsx` — the single source of truth that renders the RAC `<Text slot="description">` and `<FieldError>` slots.
- Props added to: `TextField`, `NumberField`, `SearchField`, `Select`, `ComboBox`, `RadioGroup`, `CheckboxGroup`, `DateField`, `TimeField`.
- Single controls (`Checkbox`, `Switch`, `Radio`, `Label`) intentionally **not** changed — per RAC/React Spectrum convention, helper text belongs at the group level.
- Code Connect (`.figma.tsx`) mappings updated to map Figma's `description` to the new prop.
- `TextField` story updated (prop + back-compat slot story) with a `play` test asserting render + `aria-describedby` wiring.
- `minor` changeset (additive, non-breaking — existing slotted children keep working).

## Review focus

**Please review `TextField` first** — it is the reference implementation. Scrutinize:
- `src/fieldHelpers.tsx` (the shared helper — the only real logic)
- `src/TextField.tsx`
- `figma/TextField.figma.tsx`

The other 8 components apply the **identical** mechanical change: destructure `description, errorMessage` out of props and route children through `renderFieldHelpers`. Once the `TextField` approach is approved, the rest are replicas.

## Test plan

- [ ] `pnpm --filter @launchpad-ui/components test` — run the `TextField` `play` test
- [ ] Storybook: verify the `TextField` `Example` renders the description and `DescriptionAsSlot` still works
- [ ] `figma connect publish` validation — confirm the Figma `Description` property/layer names in the updated `.figma.tsx` files match the actual Figma components (several were added following the existing `'Description?'` / `'Description'` convention and need verification against the live file)

[DSYS-194]: https://launchdarkly.atlassian.net/browse/DSYS-194?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

<a href="https://app.devin.ai/review/launchdarkly/launchpad-ui/pull/1941" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/packages/components/src/fieldHelpers.tsx
+++ b/packages/components/src/fieldHelpers.tsx
@@ -1,0 +1,25 @@
+import type { ReactNode } from 'react';
+
+import { FieldError } from './FieldError';
+import { Text } from './Text';
+
+interface FieldHelperProps {
+	/** Helper text describing the field, rendered below the input. */
+	description?: ReactNode;
+	/** Error message shown when the field is invalid. */
+	errorMessage?: ReactNode;
+}
+
+const renderFieldHelpers = (
+	children: ReactNode,
+	{ description, errorMessage }: FieldHelperProps,
+) => (
+	<>
+		{children}
+		{description != null && <Text slot="description">{description}</Text>}
+		{errorMessage != null && <FieldError>{errorMessage}</FieldError>}
+	</>
+);
+
+export { renderFieldHelpers };
+export type { FieldHelperProps };
--- a/.changeset/field-description-prop.md
+++ b/.changeset/field-description-prop.md
@@ -1,0 +1,7 @@
+---
+"@launchpad-ui/components": minor
+---
+
+Add ergonomic `description` and `errorMessage` props to field components: `TextField`, `NumberField`, `SearchField`, `Select`, `ComboBox`, `RadioGroup`, `CheckboxGroup`, `DateField`, and `TimeField`.
+
+These render the React Aria `description` slot (`<Text slot="description">`) and `FieldError` internally, matching the Figma library's flat `description` property and resolving the Code Connect drift. The change is additive and non-breaking: passing helper text as `<Text slot="description">` / `<FieldError>` children remains fully supported.
--- a/packages/components/figma/CheckboxGroup.figma.tsx
+++ b/packages/components/figma/CheckboxGroup.figma.tsx
@@ -25,7 +25,7 @@ figma.connect(
 		example: (props) => (
 			<>
 				{props.label}
-				<CheckboxGroup>{props.children}</CheckboxGroup>
+				<CheckboxGroup description={props.description}>{props.children}</CheckboxGroup>
 			</>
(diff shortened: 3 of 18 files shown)


packages/components/src/fieldHelpers.tsx
```diff
@@ -0,0 +1,25 @@
...
+const renderFieldHelpers = (
+	children: ReactNode,
+	{ description, errorMessage }: FieldHelperProps,
+) => (
+	<>
+		{children}
+		{description != null && <Text slot="description">{description}</Text>}
+		{errorMessage != null && <FieldError>{errorMessage}</FieldError>}
+	</>
+);
```
nhironaka, line 22: @apucacao @zakk-verrilli-ld 
we've always approach launchpad component in composable architecture, and this approach seems to go against that. 

do we want to expose description/error as props? or just direct develops to use the slot to render error/descriptions?

apucacao: Seems simpler to me do use the `slot` API, but I don't have much context around this change. Is this just about ergonomics? Agents and humans seem to deal with slots fine, and — assuming it is mostly about ergonomics — I'd argue adding another mechanism of providing values for these slots could add confusion: 2 ways of doing the exact same thing?

zakk-verrilli-ld: Agreed slots are the way. Both `description` and `errorMessage` are (in some cases the exclusive) slots for field components already.

[RAC Source](https://github.com/adobe/react-spectrum/blob/main/packages/react-aria-components/src/RadioGroup.tsx#L324-L325)

Is the issue here that Figma Code Connect is not wired up to generate the subcomponents? [Slack Thread](https://launchdarkly.slack.com/archives/C0B7W15FX0C/p1782402601831409)

ShreyasGit51283:
@zakk-verrilli-ld reccomends we should explore fixing the Code connect generator. Closing this PR to explore that route

PR #1951, opened 2026-07-20
feat: remove no-op --lp-spacing-100 and --lp-size-0 tokens

ld-ahartmann:
## Summary

Removes two design tokens that both resolved to `0`:

- `--lp-size-0` (`size.0` = `0`)
- `--lp-spacing-100` (`spacing.100` = `{size.0}` = `0`)

The `spacing-100` name in particular is misleading — it implies "a tiny amount of spacing" when it actually means "none at all" (see the #proj-launchpad-guild thread). Consumers should just use a literal `0`.

### Not as isolated as it looks — `size.0` had two other dependents

`size.0` wasn't only aliased by `spacing.100`; it was also referenced by two other semantic tokens. Those are repointed to a literal `0` so their **computed values are unchanged**:

- `borderWidth.100` (`--lp-border-width-100`): `{size.0}` → `0`
- `viewport.mobile` (`--lp-viewport-mobile`): `{size.0}` → `0`

(If the guild would rather keep `--lp-size-0` as the zero primitive and only drop `--lp-spacing-100`, that's an easy trim — this is a draft.)

### Changes

- `tokens/spacing.json`: drop `100`
- `tokens/size.json`: drop `0`
- `tokens/border.json`: `borderWidth.100` → literal `0`
- `tokens/viewport.json`: `viewport.mobile` → literal `0`
- `packages/components/src/styles/{Tree,TagGroup,Code}.module.css`: the design system's own 3 usages of `--lp-spacing-100` → `0`
- changeset (`tokens` minor, `components` patch)

### Rollout / coordination

The Figma sync (`sync-figma.yml`) is one-way and CREATE/UPDATE-only — it never DELETEs — so removing these from the JSON does **not** delete the corresponding Figma variables; they just stop being managed (someone can remove them in Figma manually if desired). No Figma designs break.

Other org repos still reference `--lp-spacing-100` (e.g. `launchdarkly-toolbar`'s `ShareStatePopover.css.ts` uses `var(--lp-spacing-100, 4px)` — note that once the token is gone, that fallback would flip its gap from `0` to `4px`). The gonfalon consumer-side cleanup + a lint guard is in launchdarkly/gonfalon#67565. This PR should merge only once remaining consumers are migrated.

No visual change — every affected token computed to `0` before and after.

## Testing approaches

- `pnpm --filter @launchpad-ui/tokens build` — builds clean; generated `index.css` no longer defines `--lp-size-0`/`--lp-spacing-100`, and `--lp-border-width-100: 0` / `--lp-viewport-mobile: 0` as expected. (Pre-existing "filtered out token references" warning is unrelated — present on `main`.)
- `pnpm --filter @launchpad-ui/tokens test` — 20/20 pass.
- `biome check` clean on all changed files.

Link to Devin session: https://app.devin.ai/sessions/5788404a63984bcfa815eab1f6a7d802
Requested by: @ld-ahartmann


ld-ahartmann:
closing for now.

PR #1954, opened 2026-07-22
fix(components): lock nested scroll parent when Popover is open

joannerd:
## Summary

React Aria's built-in scroll lock for modal overlays only targets `document.documentElement` (`usePreventScroll` sets `overflow: hidden` on `<html>`). Apps that scroll inside a **nested container** — e.g. an app shell where `<html>`/`<body>` are `overflow: hidden` and layout elements have `overflow: auto` — get no effective lock: the page keeps scrolling behind an open `Popover`, `Select`, `Menu`, `ComboBox`, etc. This is the reported behavior in Gonfalon's release builder, where `Select` dropdowns don't lock the page the way the React Aria docs demo does.

This adds a **generic** scroll-parent lock at the `Popover` layer (the shared overlay all of the above compose), complementing React Aria's existing document lock:

- New internal hook `usePreventScrollParent({ triggerRef, isOpen, isDisabled })`.
- On open (modal only), it walks up from the **trigger** (not the portaled popover DOM) and locks **every scrollable ancestor** between the trigger and the document root: `overflow: hidden` plus scrollbar compensation (`scrollbar-gutter: stable`, falling back to `padding-right`).
- An ancestor is matched by its `overflow` being scrollable **regardless of whether it currently overflows** — this mirrors React Aria's own unconditional `<html>` lock, and content can grow to overflow while the overlay is open. App shells commonly nest several scroll containers, so all of them are locked.
- **No-op when the document itself is the scroller** — apps already handled by React Aria are unaffected — and **skipped for non-modal popovers** (mirrors `usePreventScroll`'s `isDisabled`).
- **Reference-counted** by element via a module-level map, so nested/simultaneous overlays sharing scroll containers lock once and only restore the original inline styles when the last overlay closes. The search treats an already-locked element as a valid target (its `overflow` is now `hidden`, which `isScrollable` would otherwise reject).

`Popover` resolves `triggerRef`/`isOpen` from props first, then React Aria's `PopoverContext` / `OverlayTriggerStateContext` (how `DialogTrigger`, `Select`, etc. pass them), and installs the hook.

```tsx
// Popover.tsx (wiring)
const triggerRef = props.triggerRef ?? getTriggerRef(useContext(AriaPopoverContext));
const isOpen = props.isOpen ?? useContext(OverlayTriggerStateContext)?.isOpen ?? false;
usePreventScrollParent({ triggerRef, isOpen, isDisabled: props.isNonModal ?? false });
```

`patch`: no public API change — additive internal behavior on `Popover`.

### Why the first iteration didn't work

The initial version only locked the trigger's nearest ancestor that was **currently overflowing** (`isScrollable(node, true)`). In Gonfalon's main workspace the nested `overflow: auto` layout containers frequently are **not** overflowing at the moment the dropdown opens (their `scrollHeight === clientHeight`), so the hook found nothing and locked nothing — even though the page could still scroll. Locking every scrollable ancestor unconditionally (matching React Aria's document-level lock) fixes this.

No visual diff (scroll-locking behavior only). Verified live against a Gonfalon flag targeting page with a build of this branch consumed locally, exercising a `Select` in the same main-workspace scroll container as `ReleaseBuilderContextKindSelect`:

- Before open: the three nested `overflow: auto` layout containers (`ProjectEntityLayout__main`, `ProjectEntityLayout__workspace`, `LayoutWithSidebar__workspace`) had no inline `overflow`.
- During open: all three had inline `overflow: hidden`.
- After close: all three restored to their original (empty) inline `overflow`.

## Testing approaches

- Unit tests in `packages/components/__tests__/usePreventScrollParent.spec.tsx`:
  - locks the trigger's scroll parent while open and restores on close;
  - **locks a scroll parent that is not currently overflowing** (the case that was missed before);
  - **locks every nested scroll container** between the trigger and the document;
  - does not lock when disabled (non-modal);
  - reference-counts nested overlays sharing a scroll parent (stays locked until the last closes).
- `pnpm typecheck`, Biome, and the `usePreventScrollParent` suite pass; `@launchpad-ui/components` builds.
- Manual live verification in Gonfalon as described above.

## Notes / limitations

- Ships as a separate LaunchPad release + version bump in Gonfalon; nothing lands in Gonfalon from this PR.
- Left as **draft** for a human to review and mark ready.

Link to Devin session: https://app.devin.ai/sessions/8a5d2d3d589541a9b4bf71891c8017a4
Requested by: @joannerd

### DIFF
--- a/.changeset/popover-scroll-parent-lock.md
+++ b/.changeset/popover-scroll-parent-lock.md
@@ -2,4 +2,4 @@
 '@launchpad-ui/components': patch
 ---
 
-`Popover` now also prevents scrolling of the trigger's nearest scrollable ancestor while open. React Aria's built-in scroll lock only targets `document.documentElement`, so in apps that scroll inside a nested container (e.g. an app shell with `overflow: auto` on a layout element) the page still scrolled behind an open `Popover`, `Select`, `Menu`, etc. The lock is a no-op when the document itself is the scroller and is skipped for non-modal popovers.
+`Popover` now also prevents scrolling of the trigger's scrollable ancestors while open. React Aria's built-in scroll lock only targets `document.documentElement`, so in apps that scroll inside nested containers (e.g. an app shell with `overflow: auto` on layout elements) the page still scrolled behind an open `Popover`, `Select`, `Menu`, etc. The lock now applies to every scrollable ancestor of the trigger regardless of whether it currently overflows, mirroring the document-level lock. It is a no-op when the document itself is the scroller and is skipped for non-modal popovers.
--- a/.changeset/popover-scroll-parent-lock.md
+++ b/.changeset/popover-scroll-parent-lock.md
@@ -1,0 +1,5 @@
+---
+'@launchpad-ui/components': patch
+---
+
+`Popover` now also prevents scrolling of the trigger's nearest scrollable ancestor while open. React Aria's built-in scroll lock only targets `document.documentElement`, so in apps that scroll inside a nested container (e.g. an app shell with `overflow: auto` on a layout element) the page still scrolled behind an open `Popover`, `Select`, `Menu`, etc. The lock is a no-op when the document itself is the scroller and is skipped for non-modal popovers.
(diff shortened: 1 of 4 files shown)


joannerd:
Closing this PR as this approach doesn't work because the issue is in the parent component(s).

PR #1961, opened 2026-07-28
docs(components): add Forms guidelines page to Storybook

cmwinters:
## Summary

Adds a **Forms overview page at the root of the `Components/Forms` group** — a home for written guidance, which the design system didn't have. Storybook carried API references (autodocs) and live examples (stories, recipes), but nowhere to say "here is how to compose a form."

Clicking **Forms** in the sidebar now opens this overview; expanding it still lists the ten form components. A docs entry whose title matches a folder path coexists with that folder's children without conflict, so this needed **zero config change**.

Files sit alongside the stories they document — `Forms.mdx` and `Forms.stories.tsx` next to `Form.stories.tsx`. Their header comment flags that these are cross-cutting form examples rather than the `Form` component's own stories, since the singular/plural filenames are easy to confuse.

### The main thing this page fixes

**No form component accepts `label`, `description`, `errorMessage` or `placeholder` as a prop.** React Aria deliberately `Omit`s them — they're React Spectrum props. The real pattern is slot composition:

```tsx
<TextField isRequired>
	<Label>Email</Label>
	<Input placeholder="Enter a value" />   {/* placeholder is on Input */}
	<Text slot="description">Helper text</Text>
	<FieldError />
</TextField>
```

This is easy to get wrong because `AGENTS.md:59` mentions `description`/`errorMessage` as *the naming convention to mirror if you ever add such a prop* — not as a description of today's API. The page quotes the `.d.ts` `Omit` directly so it can't drift.

### Three gaps recorded rather than papered over

1. **Required fields render no visible marker.** `isRequired` sets `aria-required` and native `required`, but `Label` has no `isRequired` awareness. The deprecated `@launchpad-ui/form` had `RequiredAsterisk` and an `(optional)` suffix — neither has a modern equivalent.
2. **`FieldError` has no icon and no live region.** Styled text only; the deprecated package had an alert icon with `aria-live="polite"`.
3. **Helper text fails colour contrast in picker items** — `Select`/`ComboBox` stories suppress the axe `color-contrast` rule for `[slot="description"]` rather than fixing the token.

These are arguably the strongest argument for the page: they're invisible today. Each deserves its own ticket.

### Content guidelines are a scaffold with a named gap

Per the agreed approach, wording rules are **not** invented here. I searched exhaustively — the repo has **no** written copy standards: zero hits for "copy standard", "sentence case", "title case", "content guideline", "microcopy", "voice", "tone", "capitalization" across all markdown, MDX and code. Commit `157f6971` ("follow our copy standards") is a **one-line story edit** (`"Secondary action"` → `"Cancel"`).

So that section lists the questions a content designer needs to answer, plus the de facto conventions observable in our stories as raw material. **It needs an owner** — either someone supplies a canonical LD content-standard URL to link, or this page becomes the first written version.

⚠️ **Not attached — I have no browser tooling in this environment.** Verified structurally instead (below), but **the visual pass is genuinely unverified**: reading the page in dark mode, and confirming each `<Canvas>` renders live rather than an error placeholder. Please eyeball those, or the Chromatic snapshots will.

## Testing approaches

Verified against the built Storybook output rather than by eye:

- `pnpm storybook:build` exits 0 — the MDX compiles
- `index.json` shows `components-forms--docs` with `title: 'Components/Forms'`, type `docs`, tag `unattached-mdx`
- That entry **coexists with all 44 `Components/Forms/*` child entries** with no duplicate-title conflict — verified by rebuilding after the retitle
- The companion story file leaks **no** sidebar entries — its stories resolve to tags `['test']` with `dev` and `autodocs` stripped, exactly matching how `recipes/composition.stories.tsx` indexes
- The compiled docs chunk references `forms.stories` and all five story symbols, so the `<Canvas of={...}>` imports bundle correctly
- `toc:!0` present in the chunk — the `docs.toc` parameter does take effect on an unattached MDX page (this was the uncertain bit; nothing else in the repo sets it)
- `pnpm typecheck` and `pnpm check` clean
- Slot usage cross-checked against existing `CheckboxGroup`/`RadioGroup` stories, since a slot mismatch throws at render (see DSYS-157)

Examples are written as **real stories, not fenced code blocks**, so they're type-checked and can't drift from the component APIs. Given the `label`/`description` confusion above, a wrong sample would be worse than no page.

## Follow-ups worth filing

Found while auditing, all out of scope here:

- `RadioButton.tsx:19-23` has **Radio's JSDoc copy-pasted** ("A radio represents an individual option…") — it renders as a button. `TextArea.tsx:23-27` typo "mult-line". `ComboBoxClearButton`, `CheckboxIcon`, `RadioIcon` exported with no JSDoc. Autodocs is global, so these are public today.
- No story files for `TextArea`, `Radio`, `RadioButton`, `Input`, `FieldError`.
- **Dead CSS**: `Label.module.css:13-18` keys on a bare `orientation` attribute but `Form.tsx:38` emits `data-orientation`. `Form`'s horizontal `div[data-rac]` selector also can't match `Checkbox`/`Radio`/`Switch` (root `<label>`).
- **Broken axe suppression**: `ComboBox.stories.tsx:172-178` uses `'*not([aria-hidden]...)'` — missing the `:` before `not(`, so it likely isn't applying.
- **ADR-001 and ADR-007 are stale** — both describe Playwright a11y testing, but there's no Playwright dependency or config in the repo and no `axe` import in any test. Real coverage is the addon-a11y panel plus Chromatic.
- **Nine of twelve `Legacy/Form/*` stories are untagged**, so deprecated form components look like live docs — actively confusing next to a new Forms guidelines page.

### DIFF
--- a/.storybook/main.ts
+++ b/.storybook/main.ts
@@ -1,6 +1,8 @@
 import type { StorybookConfig } from '@storybook/react-vite';
 import type { UserConfig } from 'vite';
 
+import remarkGfm from 'remark-gfm';
+
 const config: StorybookConfig = {
 	stories: [
 		'../docs/stories/**/*.mdx',
@@ -9,7 +11,18 @@ const config: StorybookConfig = {
 	],
 	addons: [
 		'@storybook/addon-a11y',
-		'@storybook/addon-docs',
+		{
+			name: '@storybook/addon-docs',
+			options: {
+				// MDX only implements CommonMark, so GFM syntax — tables, strikethrough,
+				// task lists, bare-URL autolinks — renders as literal text without this.
+				mdxPluginOptions: {
+					mdxCompileOptions: {
+						remarkPlugins: [remarkGfm],
+					},
+				},
+			},
+		},
 		'@storybook/addon-designs',
 		'@storybook/addon-themes',
 		'storybook-addon-pseudo-states',
--- a/.storybook/manager.css
+++ b/.storybook/manager.css
@@ -118,6 +118,42 @@
 	font-weight: 400 !important; /* matching --lp-font-weight-regular */
 }
 
+/* Prose tables in MDX docs pages. GFM tables (enabled via remark-gfm in main.ts)
+   render as bare <table> elements that Storybook does not style at all — only
+   .docblock-argstable is themed, elsewhere in this file.
+
+   Deliberately scoped to DIRECT children of .sbdocs-content: MDX renders top-level
+   prose elements there, while a <Canvas> nests story content deeper. That keeps
+   these rules off tables rendered by the Table component's own stories.
+
+   These use --lp-* tokens rather than this file's hardcoded-hex convention because
+   .sbdocs only exists in the preview frame, where the token CSS is loaded and
+   [data-theme] flips the values for dark mode. Text color is deliberately left to
+   the .sbdocs prose rules below, which set it with !important for both themes.
+
+   The .sbdocs.sbdocs-content compound matches the docs container (which carries both
+   classes) and keeps specificity level with the argstable rules above, so this does
+   not read as a descending-specificity conflict. */
+.sbdocs.sbdocs-content > table {
+	border-collapse: collapse;
+	width: 100%;
+	margin: var(--lp-spacing-400) 0;
+	font: var(--lp-text-body-2-regular);
+}
+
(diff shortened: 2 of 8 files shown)


cmwinters:
Closing for now, until we figure out how we want to approach this

PR #1975, opened 2026-08-04
chore: pin dependency versions exactly

apucacao:
## Summary

Gonfalon requires every dependency version in package.json to be exact (no `^` or `~` ranges), so our ranges would fail its checks the moment this code moves there. This applies the same policy here first, where our own CI can prove nothing changes.

- Pinned 38 devDependencies to the exact versions that were already installed. Nothing was upgraded or downgraded.
- Regular dependencies of published packages were already exact, so nothing to do there.
- Peer dependencies keep their ranges on purpose — ranges there are what consuming apps rely on (including react-router's `^7.15.1 || ^8.2.0`).
- A lint rule now rejects any new range, so this can't quietly regress. Verified: reintroducing a `^` fails the check.
- The VS Code extension is excluded and untouched: pinning its `@types/vscode` breaks its packaging validation, and we're deliberately leaving that app alone.
- The lockfile shrank slightly because removing ranges let the installer collapse a duplicated set of packages. No installed version changed, and repeat installs are stable.

Builds on #1973.

n/a

## Testing approaches

Build, all 273 tests, typecheck, lint, and format checks pass. A fresh install changes nothing.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Low Risk**
> Tooling and lockfile specifier changes only; published runtime dependencies and consumer peer ranges are unchanged.
> 
> **Overview**
> Aligns the monorepo with **gonfalon-style fixed dependency versions** by replacing `^`/`~` on **devDependencies** (and internal app deps) with **exact** specifiers that match what was already installed—no intentional upgrades or downgrades.
> 
> **Enforcement:** `.syncpackrc` adds `semverGroups` requiring empty ranges (`""`) for prod/dev on external deps (`!@launchpad-ui/**`), while **ignoring** `launchpad-design-system` (VS Code extension) and keeping **workspace** `@launchpad-ui/**` on `workspace:~`. Existing `postinstall` `syncpack lint` will reject reintroduced ranges.
> 
> **Scope:** Published package **runtime `dependencies`** were already exact; **peerDependencies** still use ranges for consumers. `pnpm-lock.yaml` updates specifiers only (lockfile may shrink from deduping). `packages/tokens/package.json` field order changes are from syncpack sorting, not functional edits.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b8dfbca12923aea96cecbbbc8a46ef097c25bda3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


.syncpackrc
```diff
@@ -2,6 +2,30 @@
   "$schema": "./node_modules/syncpack/schema.json",
   "sortPackages": true,
   "strict": true,
+  "semverGroups": [
+    {
+      "label": "VS Code extension is intentionally left unpinned (engines.vscode compatibility)",
```
apucacao, line 7: Not sure if I mentioned this elsewhere but I'm not worrying about the vscode extension at all in the migration. It's old, and it's unclear how many consumers we have. So for now,  I mostly ignore it to unblock myself.

nhironaka (approved):

PR #1996, opened 2026-08-12
feat(components): add labelPosition to Switch

hsadhvani:
## Summary

`Switch` renders `[track][children]`, so a visible label always sits after the toggle and there's no way to put the label first without leaving the switch. Consumers that want label-before-toggle currently either override `flex-direction` locally (see launchdarkly/gonfalon#69412) or render the text as a sibling element with `aria-label` on the switch — the latter drops the text out of the `<label>`, so clicking the words no longer toggles and the visible text is no longer the accessible name.

This adds a `labelPosition` variant, defaulting to `end` (current behavior):

```jsx
<Switch labelPosition="start">AI assist</Switch>
```

```css
.switch.labelStart {
  flex-direction: row-reverse;
}
```

The label stays `children` of the switch, so it remains inside the `<label>` — clickable and still the accessible name — and only the visual order changes. Track internals (On/Off labels, handle translation, compact) are untouched.

Naming: RAC's `Switch` has no equivalent prop, so this follows React Spectrum's `labelPosition` name with CSS-logical `start`/`end` values rather than inventing something new.

New Storybook story `Components/Forms/Switch → LabelStart` covers it (Chromatic will diff it); to follow.

## Testing approaches

- `pnpm vitest run packages/components/__tests__/Switch.spec.tsx` — 7 passed, including new cases for the default (label after track) and `labelPosition="start"`.
- `pnpm typecheck`, `pnpm oxlint:js:path` on the changed files, `pnpm fmt:check`, `pnpm lint:css:path packages/components/src/styles/Switch.module.css`.
- Changeset included (`minor` on `@launchpad-ui/components`).

Follow-up: once this is released, gonfalon can drop its local `row-reverse` override in favor of the prop.

Link to Devin session: https://app.devin.ai/sessions/6458b07f375e44219f97c3a2beb73491
Requested by: @hsadhvani

---

> [!NOTE]
> **Overview**
> Adds **`labelPosition`** (`start` | `end`, default **`end`**) on **`Switch`** so visible **`children`** can render before the track without leaving the `<label>`.
> 
> Layout uses a **`labelStart`** class that sets **`order: 1`** on the track (not **`row-reverse`**), keeping the label clickable and the accessible name aligned with the visible text. On/Off track internals are unchanged.
> 
> Tests, Storybook stories, and a **minor** changeset for **`@launchpad-ui/components`** are included.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2fa94048e556e276b0962f61c4380eb24e199492. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/switch-label-position.md
+++ b/.changeset/switch-label-position.md
@@ -1,0 +1,5 @@
+---
+'@launchpad-ui/components': minor
+---
+
+Add `labelPosition` to `Switch`. Pass `labelPosition="start"` to render a visible label before the track instead of after it, while keeping the label inside the switch so it stays clickable and remains the accessible name. Defaults to `"end"`, matching current behavior.
--- a/packages/components/__tests__/Switch.spec.tsx
+++ b/packages/components/__tests__/Switch.spec.tsx
@@ -25,6 +25,16 @@ describe('Switch', () => {
 		expect(screen.queryByText('On')).not.toBeInTheDocument();
 	});
 
+	it('renders a visible label after the track by default', () => {
+		render(<Switch>Dark mode</Switch>);
+		expect(screen.getByRole('switch', { name: /Dark mode/ }).closest('label')?.className).not.toContain('labelStart');
+	});
+
+	it('renders a visible label before the track when labelPosition is start', () => {
+		render(<Switch labelPosition="start">Dark mode</Switch>);
+		expect(screen.getByRole('switch', { name: /Dark mode/ }).closest('label')?.className).toContain('labelStart');
+	});
+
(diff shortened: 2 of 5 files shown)


ld-ahartmann (approved):

erangeles:
@hsadhvani I was wondering if we could we support this through composition rather than adding a `labelPosition` prop? For example, could the label and switch control be exposed as composable parts so consumers can choose their order while preserving the accessible labeling relationship?

PR #2018, opened 2026-08-28
feat(components): add medium and small sizes to EmptyState

ShreyasGit51283:
## Summary

`EmptyState` only had a `large` size, so empty states inside cards, panels, and table bodies had to be hand-built. This adds the `medium` and `small` sizes from the [Figma spec](https://www.figma.com/design/98HKKXL2dTle29ikJ3tzk7/branch/HBR47MxC7pZofbcHUywTDs/%F0%9F%9A%80-LaunchPad?node-id=29527-96176).

`EmptyState` now scales its `BadgeIcon`, heading, description, and buttons together from a single `size` prop, so you set `size` on `EmptyState` rather than on each child. To make that work, `BadgeIcon` gains a `BadgeIconContext` that mirrors the existing `IconContext` — the context value wins over the prop, so the badge cannot drift out of scale with its container. Existing `<BadgeIcon size="large">` calls inside an `EmptyState` keep working; the prop is just redundant now, and it has been dropped from the stories and tests.

Only `large` carries padding. `medium` and `small` leave it to the surrounding card, panel, or table body, since that is what controls density at those sizes.

In the CSS module the sizes are driven by four local custom properties (`--empty-state-gap`, `--empty-state-content-gap`, `--empty-state-heading-font`, `--empty-state-description-font`) declared on `.base` with large's values and overridden by `.medium` / `.small`, following the `--alert-color-*` pattern already in `Alert.module.css`.

Rendered in Storybook and measured against the Figma spec. Every value matches:

| | Large | Medium | Small |
|---|---|---|---|
| BadgeIcon | 64 | 36 | 24 |
| Heading | 20/28 600 | 15/24 600 | 13/20 600 |
| Description | 15/24 400 | 13/20 400 | 11/16 400 |
| Button height | 40 | 32 | 24 |
| Outer gap | 24 | 16 | 12 |
| Heading to description | 8 | 4 | 4 |

See `Recipes/EmptyState` in Storybook for the new `Large`, `Medium`, and `Small` stories.

## Testing approaches

- Four new cases in `packages/components/__tests__/EmptyState.spec.tsx`: a parameterised check that each size applies the right root, badge, and button classes, plus one asserting that `EmptyState` overrides a `size` set directly on `BadgeIcon`. Full `packages/components` and `packages/icons` suites pass (131 tests).
- Verified the rendered geometry in a running Storybook by reading computed styles, which produced the table above.
- `pnpm typecheck`, `stylelint`, `oxlint`, and `oxfmt --check` all pass.

## Notes for the reviewer

Code Connect is repointed to node `29527-96176` and now maps the `Size` variant. That variant set currently only exists on the `HBR47MxC7pZofbcHUywTDs` Figma branch, so the mapping will not resolve until that branch merges.

### DIFF
--- a/.changeset/empty-state-sizes.md
+++ b/.changeset/empty-state-sizes.md
@@ -1,0 +1,6 @@
+---
+'@launchpad-ui/components': minor
+'@launchpad-ui/icons': minor
+---
+
+Add `medium` and `small` sizes to `EmptyState`. `EmptyState` now scales its `BadgeIcon`, heading, description, and buttons together, so set `size` on `EmptyState` rather than on the children. Use `large` for full-page and main-content empty states, `medium` inside cards, panels, and table bodies, and `small` in dense containers. Only `large` has padding of its own; `medium` and `small` leave it to the surrounding container. `BadgeIcon` gains a `BadgeIconContext` so a composing component can drive its size.
--- a/packages/components/__tests__/EmptyState.spec.tsx
+++ b/packages/components/__tests__/EmptyState.spec.tsx
@@ -5,13 +5,15 @@ import { render, screen } from '@launchpad-ui/test-utils';
 
 import { Button, ButtonGroup, EmptyState, Heading, Text } from '../src';
 
+import badgeIconStyles from '../../icons/src/styles/BadgeIcon.module.css';
+import buttonStyles from '../src/styles/Button.module.css';
 import emptyStateStyles from '../src/styles/EmptyState.module.css';
 
 describe('EmptyState', () => {
 	it('renders heading and description', () => {
 		render(
 			<EmptyState>
-				<BadgeIcon size="large" aria-hidden>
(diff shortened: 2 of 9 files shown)


ShreyasGit51283:
Closing in favour of #2019.

This PR was built against Figma component set `29527:96176`, which has since been rebuilt as `29920:114`. Enough changed that a clean PR reads better than a series of corrections on top of this one:

- The medium and small headings moved from semibold to the medium weight.
- Block spacing and the heading-to-description gap are now uniform across all three sizes (24 and 8) rather than scaling down, so this PR renders medium and small 20px and 28px shorter than the design.
- The variant property was renamed and its options changed, and `Border?`, `Icon`, and `Button label` no longer exist on the component, so the Code Connect mapping here points at a deleted node.

#2019 carries the same `BadgeIconContext` approach and the same tests, rebased onto `main`.

PR #2019, opened 2026-08-31
feat(components): add medium and small sizes to EmptyState

ShreyasGit51283:
Replaces #2018, which was built against an earlier revision of the Figma component. That component set was rebuilt, so this is a clean PR against the current design rather than a series of corrections.

## Summary

`EmptyState` only had a `large` size, so empty states inside cards, panels, and table bodies had to be hand-built. This adds `medium` and `small` from the [current Figma component set](https://www.figma.com/design/98HKKXL2dTle29ikJ3tzk7/%F0%9F%9A%80-LaunchPad?node-id=29920-114).

`EmptyState` now scales its `BadgeIcon`, heading, description, and buttons together from a single `size` prop, so you set `size` on `EmptyState` rather than on each child. To make that work, `BadgeIcon` gains a `BadgeIconContext` that mirrors the existing `IconContext` — the context value wins over the prop, so the badge cannot drift out of scale with its container. Existing `<BadgeIcon size="large">` calls inside an `EmptyState` keep working; the prop is just redundant now, and it has been dropped from the stories and tests.

Spacing is uniform across all three sizes in the design (24px between blocks, 8px between heading and description), so only the type scale and the control sizes vary. That keeps the CSS small: `.medium` and `.small` each override just two custom properties for the heading and description type.

Only `large` carries padding. `medium` and `small` leave it to the surrounding card, panel, or table body, since that is what controls density at those sizes.

Rendered in Storybook and asserted against the Figma spec. All 21 measurements match:

| | Large | Medium | Small |
|---|---|---|---|
| BadgeIcon | 64 | 36 | 24 |
| Heading | 20/28 600 | 15/24 500 | 13/20 500 |
| Description | 15/24 400 | 13/20 400 | 11/16 400 |
| Button height | 40 | 32 | 24 |
| Block spacing | 24 | 24 | 24 |
| Heading to description | 8 | 8 | 8 |

See `Recipes/EmptyState` in Storybook for the new `Large`, `Medium`, and `Small` stories.

## Testing approaches

- Four new cases in `packages/components/__tests__/EmptyState.spec.tsx`: a parameterised check that each size applies the right root, badge, and button classes, plus one asserting that `EmptyState` overrides a `size` set directly on `BadgeIcon`. Full `packages/components` and `packages/icons` suites pass (123 tests).
- Verified the rendered geometry against the Figma values by reading computed styles in a running Storybook, which produced the table above.
- `tsc --noEmit`, `stylelint`, `oxlint`, and `oxfmt --check` all pass.

## Notes for the reviewer

**Code Connect is narrower than before, deliberately.** The rebuilt component set only exposes `Heading`, `Description`, and the variant property. Three properties the old mapping used — `Border?`, `Icon`, and `Button label` — no longer exist on it; the latter two moved down into the nested Badge Icon and Button instances. Reaching them again needs `figma.nestedProps`, which has no precedent in this repo, and `figma connect publish` only runs on push to `main`, so a bad mapping would not surface until after merge. I mapped only the properties I could verify and left the icon and button label as literals in the example. Restoring them via `figma.nestedProps` is a good follow-up once someone can validate against Figma with a token.

**Two things worth raising with the designer.** The variant property is unnamed, so it reads as `Property 1` with a `Default` option instead of `Size` / `Large`; the mapping handles it but would be less brittle if renamed. Separately, the Figma component has zero padding on every variant including `Default`, while `large` ships 28px today — that predates this PR, and removing it would visibly change every existing `EmptyState`, so I left it alone.

**Not included:** the per-variant widths (400 for Default, 320 for the others) and the 560px max-width. Fixed widths felt like the container's job, but happy to add them if you disagree.

### DIFF
--- a/.changeset/empty-state-sizes.md
+++ b/.changeset/empty-state-sizes.md
@@ -1,0 +1,6 @@
+---
+'@launchpad-ui/components': minor
+'@launchpad-ui/icons': minor
+---
+
+Add `medium` and `small` sizes to `EmptyState`. `EmptyState` now scales its `BadgeIcon`, heading, description, and buttons together, so set `size` on `EmptyState` rather than on the children. Spacing is the same at every size; only the type scale and the control sizes change. Use `large` for full-page and main-content empty states, `medium` inside cards, panels, and table bodies, and `small` in dense containers. Only `large` has padding of its own; `medium` and `small` leave it to the surrounding container. `BadgeIcon` gains a `BadgeIconContext` so a composing component can drive its size.
--- a/packages/components/__tests__/EmptyState.spec.tsx
+++ b/packages/components/__tests__/EmptyState.spec.tsx
@@ -5,13 +5,15 @@ import { render, screen } from '@launchpad-ui/test-utils';
 
 import { Button, ButtonGroup, EmptyState, Heading, Text } from '../src';
 
+import badgeIconStyles from '../../icons/src/styles/BadgeIcon.module.css';
+import buttonStyles from '../src/styles/Button.module.css';
 import emptyStateStyles from '../src/styles/EmptyState.module.css';
 
 describe('EmptyState', () => {
 	it('renders heading and description', () => {
 		render(
 			<EmptyState>
-				<BadgeIcon size="large" aria-hidden>
+				<BadgeIcon aria-hidden>
 					<Icon name="folders" />
 				</BadgeIcon>
 				<Heading>No projects yet</Heading>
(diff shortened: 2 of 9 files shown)


cmwinters (commented):
Looking good, just a few nits on spacing:

Before | After
--- | ---
 Image | Image
Image | Image

- We want to adjust the space between the `Heading` and `Text` for the medium and small variants
- Add `grid-template-rows: min-content min-content var(--lp-spacing-200) min-content min-content;` to your selectors for the medium and small sizes.
- Loom https://www.loom.com/share/ca3e3ad09521407ebb6a68738ab18a8e

Also, this PR in on LaunchPad UI which is being deprecated because [LaunchPad has moved to Gonfalon](https://github.com/launchdarkly/gonfalon/tree/main/launchpad-ui). **We shouldn't merge this.** Let's open a similar PR on gonfalon.

ShreyasGit51283:
Closing in favour of launchdarkly/gonfalon#71171, which ports this change to gonfalon now that LaunchPad lives there under `launchpad-ui/` and this repo is deprecated.

The commit was replayed as a patch, so the component change is identical to what was reviewed here. @cmwinters' spacing feedback is folded into that PR rather than left as a follow-up — the heading-to-description gap tightens to `--lp-spacing-200` (4px) at `medium` and `small`. It is implemented via a `--empty-state-heading-gap` custom property on `.base` that the smaller sizes override, matching the pattern the file already uses for the type scale, instead of repeating the full `grid-template-rows` in both selectors. Same geometry; happy to switch to the literal form there if preferred.

